PR #160, opened 2014-12-31
Conversion of gh-pages content into AsciiDoc

spohnan:
This patch does the following:

* The content from the gh-pages branch is converted to AsciiDoc and copied to the docs directory in master
* A new profile is added to the parent pom to build the documentation. This profile is not executed during the normal build and has to be explicitly reqeusted
* It _does not_ alter the current project web site content, it just stages the content needed to do so at some point in the future
* This is intended to get everything staged so we can test the creation of docs and website content before actually pushing it out to the public site

### Code to create docs in geowave/target/site
```
# Build javadocs
mvn clean install javadoc:aggregate -DskipITs=true -DskipTests=true

# Build docs
mvn -P docs -pl docs clean install
```


rfecher:
I'm going to just leave the doc changes for Chris to merge, considering he did nearly all of the first iteration of it

PR #191, opened 2015-01-22
Updated geowave-test to support testing for GEOSERVER 2.6.2.

jwomeara:
The travis tests are currently all passing, but initially they failed for the combination of Geoserver 2.6.2 and Accumulo 1.5.1.  We should keep an eye on this, I suppose...


jwomeara:
Let's not merge this in for now.  The reason the build is failing is because the WPS jars that we drop into WEB-INF lib contain Geoserver 2.7 classes and packages.  Specifically, the error that's being called out in our integration test refers to the class WpsAccessRuleDAO in the package org.geoserver.wps.security.  org.geoserver.wps.security does not exist on the Geoserver github page for the 2.6.2 tag, and it also is not included in the wps-core jar that you can download from their official website.

chrisbennight:
Issue with the artifacts in the boundless repo - looks like a build quirk, the 2.6.2 jar/war contains 2.7-beta artifacts, and vice versa - here's a listing of the extracted lib directory, filtered for versions (from the 2.6.2 jar in this case)

2 6 2-artifact

geowave-test/pom.xml
```diff
@@ -91,7 +91,7 @@
 									<type>war</type>
 									<overWrite>true</overWrite>
 									<outputDirectory>${project.build.directory}/geoserver</outputDirectory>
-									<excludes>**/commons-dbcp-1.4.jar,**/*1.7-SNAPSHOT.jar,**/*2.7-SNAPSHOT.jar,**/*20140915*.jar,**/*13-SNAPSHOT.jar</excludes>
+									<excludes>**/*1.7-beta.jar,**/*13-beta.jar,**/*2.7-beta.jar,**/commons-dbcp-1.4.jar,**/*1.7-SNAPSHOT.jar,**/*2.7-SNAPSHOT.jar,**/*20140915*.jar,**/*13-SNAPSHOT.jar</excludes>
```
chrisbennight: Probably remove this exclusion when the artifacts get fixed - shouldn't need it

jwomeara:
Actually I added exclusions for those jars in the geowave-test pom to fix
that problem. If you look at the unpacked geoserver folder in the target
directory youll see that they are properly missing. The problem now is due
to the wps jars that we add to the web-inf/lib directory. Boundless
included some classes and packages from their upcoming 2.7 release in their
wps jars (but only the ones checked into their repo; the ones on theain
site are fine).

Do you know of anyone at boundless who we can talk to to get this fixed?
It is affecting 2.5.2 as well.

geowave-test/pom.xml
jwomeara: The same goes for the snapshot jars. Those are excluded to fix 2.5.2. We
need to get in contact with boundless to have them fix their jars.

chrisbennight:
Makes complete sense why you added them - that's what I meant by "remove when the artifact gets fixed" - i.e. when we get a 2.6.2 war/jar published to the boundless repo that doesn't have 2.7 classes :) - just wanted to make sure when it was fixed (it being fixed artifacts published to the maven repo) we didn't leave that code in there

chrisbennight:
Re: getting it fixed in repo - 
http://osgeo-org.1560.x6.nabble.com/gs-web-app-2-6-2-war-jar-artifacts-contains-2-7-beta-artifacts-and-vice-versa-td5183266.html

jwomeara:
Agreed.

chrisbennight:
All add the 2.5.2 bit in there also; didn't check that one

jwomeara:
2.6.1 is the only one that seems to work without adding exclusions. I
didn't spot any differences when comparing those jars with the ones from
the main site

PR #194, opened 2015-01-23
GEOWAVE-189

rwgdrummer:


geowave-store/src/main/java/mil/nga/giat/geowave/store/query/TemporalConstraints.java
```diff
@@ -64,6 +69,45 @@ else if (nextNeighbor.getEndTime().after(
...
 
+	public TemporalConstraints upperBounds(
+			TemporalConstraints other ) {
+		final TemporalConstraints c = new TemporalConstraints();
+		c.add(new TemporalRange(
+				max(
+						this.getStartRange().getStartTime(),
+						other.getStartRange().getStartTime()),
+				max(
+						this.getStartRange().getStartTime(),
```
rfecher: this looks like a minor oversight. I think it is supposed to be this.getEndRange().getEndTime(). are these bounds(), lowerBounds(), and upperBounds() methods used?  I didn't see where at least and was just trying to understand the intended purpose particularly with lower and upper bounds methods.  I was thinking set theory operations like union and intersection could make sense on time ranges but this upper and lower bounds for time ranges that don't overlap would produce invalid time ranges, and it just wasn't clear to me what its supposed to be doing particularly in that non-overlapping case.

rwgdrummer:
Comments addressed as part of review.

PR #197, opened 2015-01-24
tweaks to make jacoco work / cleanup cobertura error spam

chrisbennight:


geowave-test/pom.xml
```diff
@@ -91,7 +92,7 @@
 									<type>war</type>
 									<overWrite>true</overWrite>
 									<outputDirectory>${project.build.directory}/geoserver</outputDirectory>
-									<excludes>**/commons-dbcp-1.4.jar,**/*1.7-SNAPSHOT.jar,**/*2.7-SNAPSHOT.jar,**/*20140915*.jar,**/*13-SNAPSHOT.jar</excludes>
+									<excludes>**/commons-dbcp-1.4.jar,**/*1.7-SNAPSHOT.jar,**/*2.7-SNAPSHOT.jar,**/*20140915*.jar,**/*13-SNAPSHOT.jar,**/slf4j*.jar</excludes>
```
chrisbennight: Had to add ,**/slf4j*.jar to the exclusions list here as older versions of geoserver were bringing in an older version that conflicted with some hadoop dependencies

.travis.yml
```diff
@@ -17,10 +17,12 @@ cache:
   directories:
   - $HOME/.m2
 install: "mvn -q clean install javadoc:aggregate -Daccumulo.version=${ACCUMULO_VERSION} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -DskipITs=true -DskipTests=true"
-script: "mvn -q verify cobertura:cobertura -Daccumulo.version=${ACCUMULO_VERSION} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION}"
+script: "mvn -q verify -Daccumulo.version=${ACCUMULO_VERSION} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION}"
 before_install:
   - export MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=192m"
   - chmod +x .utility/push-javadoc-to-gh-pages.sh
+  - mkdir mcp; cd mcp; git clone --branch=maven-munge-hack git://github.com/chrisbennight/coveralls-maven-plugin.git; cd coveralls-maven-plugin; git submodule init; git submodule update; mvn -q clean install; cd ../..;
```
chrisbennight: The combination of munge + jacoco multi-module aggregation was breaking.  Strangely only on the travis-ci build slaves, not on win 8.1 or ubuntu 14.04 LTS or OS X 10.10.   I patched a version of the coveralls-maven-plugin to not error out on this (and print out the messages to the log) - we are loosing coverage on one file (GeowaveInputFormat);  I think as soon as Accumulo releases 1.7 and we can deprecate 1.5 we can get rid of the munge support and this issue will go away.   

This fix I made isn't really appropriate to issue a pull request on, and isn't needed to build generally, just for travis to develop code coverage reports, so I figure just leave it until munge does - then this line can die as well.

PR #209, opened 2015-01-29
release v0.8.2

rfecher:


chrisbennight:
LOL :8ball:  Should I make a ticket to use maven release plugin? :)

rfecher:
i thought we had needed to put the scm tags in our parent pom...but i see that they are in there.  oh well

rfecher:
actually I had tried the release plugin last night on project and it gives the error message: "geowave-parent: No SCM URL was provided to perform the release from"  Clearly we have it in our parent POM, but I think the real issue is that we have a SNAPSHOT dependency for the coveralls hack.  I try even using mvn release:prepare and it needs to resolve snapshot dependencies and fails.  I'm going to create an issue for it.

Issue #213, opened 2015-02-03
Refactor Integration Test Environments


chrisbennight:
Integration tests environment (at least the base parts) should be moved to the geowave-test src/main package.

A program that takes a dependency on geowave-test should be able to use directly or extend one of these classes to have a fully working geowave test environment to build their integration tests against

PR #216, opened 2015-02-04
refactoring test environments into src to be reusable

chrisbennight:
Also attempt to cleanup mapreduce logspam, and add (optional) HDFS local cluster
#213

### DIFF
--- a/geowave-test/pom.xml
+++ b/geowave-test/pom.xml
@@ -45,13 +45,21 @@
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
 			<version>4.11</version>
-			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>net.lingala.zip4j</groupId>
+			<artifactId>zip4j</artifactId>
+			<version>1.3.2</version>
+		</dependency>
+		<dependency>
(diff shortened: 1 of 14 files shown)


geowave-test/src/main/java/mil/nga/giat/geowave/test/MapReduceTestEnvironment.java
```diff
@@ -1,11 +1,10 @@
-package mil.nga.giat.geowave.test.mapreduce;
+package mil.nga.giat.geowave.test;
```
rfecher: didn't like the package names?

chrisbennight:
Mea Culpa on the package change, probably a result of other operations.  If only the commit weren't squashed :)

I'll re-format with eclipse, I imported the formatter into intellij but looks like it didn't quite work.

chrisbennight:
Yeah, tested it out and formatter just isn't working right in intellij (the eclipse formatter plugin).  I added http://maven-java-formatter-plugin.googlecode.com/ to the pom, and this adds mvn java-formatter:format as a goal.  

I only reformatted geowave-tests, didn't do the rest (didn't want whitney to be suspicious!)

PR #220, opened 2015-02-06
Geowave 14

jwomeara:
Dun-dun-dunnnnn!


geowave-vector/pom.xml
```diff
@@ -67,7 +67,12 @@
 			<groupId>org.apache.curator</groupId>
 			<artifactId>curator-test</artifactId>
 			<version>2.5.0</version>
-			<scope>test</scope>
```
rfecher: ???

geowave-vector/pom.xml
```diff
@@ -67,7 +67,12 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.apache.accumulo</groupId>
```
rfecher: I'm missing something - mini accumulo at compile scope within geowave-vector?

Issue #228, opened 2015-02-06
Maven Findbugs


chrisbennight:
Add it to the project

chrisbennight:
So there's a bunch of spam changes so far  (valid things, but minimal - i.e. use Long.valueOf() instead of new Long();  iterate on the entrySet instead of the keySet if you are accessing both the key and value, always specify the charset encoding when going  to/from string<->byte[].

But there were a few good catches it made as well also, link to the comment set below.   I think it's worthwile to triage the rest and keep this as part of our build.  (down from 400ish at medium to 200ish)

chrisbennight:
I threw some comment on what I think were the worthwile catches that justify getting this running:

https://github.com/ngageoint/geowave/commit/ae42b66840fc864bc4af2c6298714bddd5ab4b28

chrisbennight:
1/2 done in the above commit, still ~ 200 issues to address.

PR #230, opened 2015-02-07
Geowave 228

chrisbennight:
#228

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
 cache:
   directories:
   - $HOME/.m2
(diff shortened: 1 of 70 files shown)


rfecher:
Rather than adding a specific constant in each module of a "CHAR_SET" can you just use StringUtils.UTF8_CHAR_SET?  It is put in geowave-index because that is the lowest level dependency and each of the other modules can use it.

chrisbennight:
That seems completely reasonable

chrisbennight:
Moved it all over except for the ingest plugins instances;  it seems reasonable that the ingest plugins might need to declare their own input encoding (though in this case they were all UTF-8)

rfecher:
fair enough, I don't think it really matters, but I didn't actually see why if the ingest-plugins were using UTF-8, that they wouldn't just re-use the constant for UTF-8 already defined in our string utilities...then again, does it really matter, no

rfecher:
haha, I was just waiting for travis to finish and I was going to merge it this morning with the ingest plugin charsets, you didn't have to change it but thanks

Issue #231, opened 2015-02-07
IT Environment fixes


chrisbennight:
The Geowave services IT fails an assertion on cleanup when no namespeaces are created.

PR #232, opened 2015-02-07
Handle no tables added case

chrisbennight:
#231


rfecher:
When you had a chance were you still planning on changing the deleteAll() operation to just throw an exception if there is an issue rather than return a flag just to make it abundantly clear?  I just don't want  to be holding up the pull request if that was not your plan (its really not a significant issue either way).

chrisbennight:
Just pushed the new method signature,  Travising right now

PR #233, opened 2015-02-09
Added a check to ensure that null iterators aren't attached to accumulo.

jwomeara:
This is a quick fix which fixes a bug we encountered during benchmarking.  Accumulo version 1.5.1 will not attach an iterator with a null configuration, but apparently 1.6.0 will.  The code was updated to avoid attaching iterators with null configurations.


jwomeara:
I need to spend some more time looking into this issue.

PR #259, opened 2015-03-12
Fixed BigDecimal readers/writers to properly serialize values, registered missing writers/readers, added integration test

chrisbennight:
Breaks BigDecimal persistence, but looks like it wasn't working before anyway


rwgdrummer:
looks good

PR #262, opened 2015-03-13
Refactor man page generation

spohnan:
This commit will fix the generation of man pages and keep the source documentation in a single location.

All man page text is now in a single file per command, with all the manpage only markup escaped so other formats can safely include the content. The escaping tokens are removed before transformation into manpages and provided as an archive to the RPM process to simplify that build a bit as well.


chrisbennight:
Looks good to me

PR #265, opened 2015-03-15
Fixes for coverity id'd issues

chrisbennight:
(don't accept this yet, wanted to open this up to discuss a few questions first)
(coverity issues can be seen @ https://scan.coverity.com/projects/3371/view_defects - login required)

### DIFF
--- a/geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/AccumuloDataStore.java
+++ b/geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/AccumuloDataStore.java
@@ -73,8 +73,6 @@ import org.apache.log4j.Logger;
 
 import com.google.common.collect.Iterators;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-
 /**
  * This is the Accumulo implementation of the data store. It requires an
  * AccumuloOperations instance that describes how to connect (read/write data)
@@ -283,8 +281,10 @@ public class AccumuloDataStore implements
 			statisticsTool.entryIngested(
 					entryInfo,
 					entry);
-			
-			statisticsTool.flush();
+
+			synchronizeStatsWithStore(
+					statisticsTool,
+					true);
 
 			return entryInfo.getRowIds();
 		}
@@ -355,7 +355,7 @@ public class AccumuloDataStore implements
 			final IngestCallback<T> ingestCallback,
 			final VisibilityWriter<T> customFieldVisibilityWriter ) {
 		if (dataWriter instanceof IndexDependentDataAdapter) {
-			ingest(
+			ingestInternal(
 					dataWriter,
 					index,
 					new IteratorWrapper<T, T>(
@@ -506,7 +506,9 @@ public class AccumuloDataStore implements
 				altIdxWriter.close();
 			}
 
-			statsCompositionTool.flush();
+			synchronizeStatsWithStore(
+					statsCompositionTool,
+					true);
 
 		}
 		catch (final TableNotFoundException | AccumuloException | AccumuloSecurityException e) {
@@ -653,9 +655,9 @@ public class AccumuloDataStore implements
 						index),
 				authorizations);
 
-		if (success) {
-			statsCompositionTool.flush();
-		}
+		synchronizeStatsWithStore(
+				statsCompositionTool,
+				success);
 		if (success && useAltIndex) {
 			deleteAltIndexEntry(
 					altIdxTableName,
@@ -663,6 +665,15 @@ public class AccumuloDataStore implements
 					adapterId);
 		}
 
+		try {
+			// issue; going to call .flush() internally even if success = false;
+			statsCompositionTool.close();
+		}
+		catch (Exception ex) {
+			LOGGER.error(
+					"Error closing statsCompositionTool",
+					ex);
+		}
 		return success;
 
 	}
@@ -902,7 +913,7 @@ public class AccumuloDataStore implements
 		return query(
 				adapter,
 				index,
-				query,				
+				query,
 				Integer.valueOf(limit),
 				null,
 				authorizations);
@@ -1132,7 +1143,7 @@ public class AccumuloDataStore implements
 				index,
 				query,
 				limit,
-				(String[])null);
+				(String[]) null);
 	}
 
 	@Override
@@ -1148,6 +1159,7 @@ public class AccumuloDataStore implements
 				null);
 	}
 
+	@Override
 	@SuppressWarnings("unchecked")
 	public <T> CloseableIterator<T> query(
 			final DataAdapter<T> adapter,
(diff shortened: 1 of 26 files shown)


geowave-store/src/main/java/mil/nga/giat/geowave/store/adapter/statistics/StatsCompositionTool.java
```diff
@@ -93,10 +97,14 @@ public void flush() {
...
-			for (final DataStatistics<T> s : statistics) {
-				statisticsStore.incorporateStatistics(s);
+
+		synchronized (MUTEX) {
+			for (final DataStatisticsBuilder<T> builder : statisticsBuilders) {
+				final Collection<DataStatistics<T>> statistics = builder.getStatistics();
+				for (final DataStatistics<T> s : statistics) {
+					statisticsStore.incorporateStatistics(s);
+				}
+				statistics.clear();
```
chrisbennight: @rfecher @rwgdrummer 
So when .flush()  (or .close()) is called on the StatsCompositionTool instance it goes through the local set of statistics, and pushes them to accumulo.

The problem is there are plenty of statistics which are non idempotent (count being one) - and I don't think there's any realistic way we can require the statistics to be idempotent (it rules it too many).

The problem is any code path which might end up calling .flush() twice  (or call a .flush() and later a .close())  will currently trigger multiple evaluations of these.

I would expect calling .flush() multiple times to be safe.

I added a statistics.clear() call here after statistics were flushed; I'm not familiar enough with the rest of the statistics stack to know if this will have other unintended side effects.  (It passes testing fine, and in a brief look seemed to be okay.)   (had to ad locking around it to ensure what was cleared was the same as what was already persisted).

geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/AccumuloDataStore.java
```diff
@@ -663,6 +664,12 @@ public boolean deleteEntry(
 					adapterId);
 		}
 
+		try {
+			//issue; going to call .flush() internally even if success = false;
+			statsCompositionTool.close();
```
chrisbennight: So StatsCompositionTool  implements AutoCloseable / has a .close() method - which to me means we should call it when the operation is done / the object goes out of scope.

Here (in the DeleteWriter) the logic seems to say that we only want to .flush() on "success";   (and .close() calls .flush()) - I think what we actually want here is a way to remove the statistic from the local queue if the operation isn't a success, no?   Leaving the StatsCompositionTool  unflushed seems like a chancy practice - depending on if it's used elsewhere it might get flushed regardless.

(didn't implement the delete/removal part, wanted to toss this out for discussion first)

geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/util/StatsTool.java
```diff
@@ -75,6 +77,12 @@ public static boolean calculateStastics(
 			}
 		}
 		statsTool.flush();
+		try {
+			statsTool.close();
```
chrisbennight: Another area that blows up logically if .flush/close aren't idempotent

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/AnalyticFeature.java
```diff
@@ -37,7 +37,10 @@ public static SimpleFeature createGeometryFeature(
 			final int zoomLevel,
 			final int iteration,
 			final long count ) {
-		assert (extraDimensionNames.length == extraDimensions.length);
+		if (extraDimensionNames.length != extraDimensions.length) {
```
chrisbennight: Style question here - 
my take is (since assert() is disabled at runtime unless you change default JVM params) we really don't want assert statements.    If they are used for argument validation then they should be moved to a pattern like this, with a useful message.  (or just removed)

That gets into what we want to set as style guidelines for the project though.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
```diff
@@ -64,52 +65,60 @@ public PropertyManagement(
...
 		final String pName = toPropertyName(propertyName);
-		if (!properties.containsKey(pName)) {
-			LOGGER.info("Setting parameter : " + pName + " to " + value.toString());
-			properties.put(
-					pName,
-					value);
-			return value;
-		}
-		return properties.get(pName);
+		synchronized (MUTEX) {
```
chrisbennight: @rwgdrummer 

All the synchronization here was because there are a few places that have the  
if (!contains) then put
pattern going on.  Definitely not an atomic operation, and it looked like this might be called from multiple threads (just at a quick glance though) - so put a lock around the pattern above + any place that was calling adding items.

Probably can be removed/unnecessary if we want to mark everything as not threadsafe. (depends on use requirements)

geowave-raster/src/main/java/mil/nga/giat/geowave/raster/RasterUtils.java
```diff
@@ -872,7 +871,7 @@ private static void create(
 					Errors.format(
 							ErrorKeys.NUMBER_OF_BANDS_MISMATCH_$3,
 							numBands,
-							min.length,
+							min == null ? -1 : min.length,
```
chrisbennight: Just a null dereference issue  on min

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
rwgdrummer: Probably cleaner to make the entire container synchronized.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/AnalyticFeature.java
rwgdrummer: That's fine. I wanted something to go 'boom' here if the invariant came up not true.

geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/util/StatsTool.java
rwgdrummer: true...just left it because the tool is hopefully temporary to something more robust and the close is the last step of the process.

geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/AccumuloDataStore.java
rwgdrummer: I do not think this is correct.  The 'close' method will not get called in this scenario and we do not want the flush if there is a failure.  
-1

Issue #271, opened 2015-03-17
JavaDoc generation fails under Java 1.8


spohnan:
The default behavior has changed under 1.8 and it now enforces more rules associated with JavaDoc markup. Long term I guess these should all be addressed, short term I'm looking to disable the new behavior so it matches 1.7.

http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html

chrisbennight:
Andy's patch patches the issue, but we should probably make a pass through our javadoc to bring it up to spec for 1.8.   I'll leave this ticket open for that.

PR #272, opened 2015-03-17
Sonar plugin version and java doc lint

spohnan:
Addresses ticket #271 which updates the sonar plugin to a version recent enough to work with the newly released Sonar v5 server and disabled JavaDoc doclint checks which now throw errors when run under Java 1.8 instead of warnings.


chrisbennight:
Looks GREAT!

PR #273, opened 2015-03-18
Geowave vagrant

jwomeara:


jwomeara:
I will merge once travis finishes.  It kind of sucks that the diff for the parent pom shows the files as being completely different...  I added one plugin and ran the formatter.  I'm not complaining about the additional lines of code though...

PR #276, opened 2015-03-18
applied eclipse formatter; attached formatter to process-sources lifecycle

chrisbennight:
As long as you actually build your code (with maven) before a push it should be formatted.


jwomeara:
ಠ_ಠ

chrisbennight:
(╯°□°）╯︵ ┻━┻

Issue #278, opened 2015-03-21
Add test cases for examples


chrisbennight:
User reports issues with exceptions in some of geowave-examples;  probably a good idea to add some tests/code coverage for those to make sure they stay in sync.

chrisbennight:
Trying to recreate issue described in #279 in the integration test

PR #281, opened 2015-03-23
Integration & unit tests for examples

chrisbennight:
* Adds geowave-examples to the pom build matrix
  * Adds unit tests for the geowave-examples instances
  * Adds integration test using the geowave-examples data set
    * Validates WMS output
    * Rendering differences between 2.5 and 2.6 cause different outputs for the same basic grid


geowave-test/src/main/java/mil/nga/giat/geowave/test/service/ServicesTestEnvironment.java
```diff
@@ -96,6 +103,7 @@ public static void startServices() {
 							":",
 							";"));
 					gsWebapp.setClassLoader(classLoader);
+					gsWebapp.setParentLoaderPriority(true);
```
chrisbennight: Ran into imageio related sealed class exceptions without this; setting this to true creates the same behaviour as tomcat.   
@spohnan should check this for our standalone version (can also be set in the WEB-INF settings file as well for jetty)

chrisbennight:
#278

jwomeara:
This is where I say, "looks good to me," right?

rfecher:
"looks good to me" (too late Whitney)

jwomeara:
Hey, what does this "revert" button do?

Issue #274, opened 2015-03-18
Reduce size of the images in the Documentation page as it takes lot of time to load


viggyprabhu:
Currently [Documentation Page](https://ngageoint.github.io/geowave/documentation.html) takes forever to load as it has lot of images which in all around 7MB together. We can try to decrease the size of the images, using sites like Tinypng.com or some other best known methods.

spohnan:
An oversight on my part that also affects the main website page. I'll get them reduced to a more reasonable size. Thanks for pointing out the problem.

chrisbennight:
probably good to do this in concern with #252 if it's still valid.

chrisbennight:
#293 Hits most of this

Issue #268, opened 2015-03-16
Documentation build configuration


spohnan:
1.) Configure the maven build so a single goal and phase will generate all content associated with site
2.) Fix bug with images includes when building from Windows

spohnan:
Docs build with images on Windows as of #434

PR #286, opened 2015-03-24
Geowave 274 image thumbnails

spohnan:
* Addresses #274 and halves the size of our documentation and site images from ~7.5MB down to ~3.5MB
* Addresses #268 and updates the build to create docs, site and javadocs in the right order when mvn install -P docs is run
* Cleaned up and updated the docs pom. Made note that due to an upstream issue PDFs will not include images when built from Windows as reported by @rwgdrummer . At the point it gets fixed upstream I'll update our plugin.
* Fixed link used to directly download an RPM from the artifact listing page


chrisbennight:
Restarted the jobs that glitched in travis, but looks good.   One small request - could you squash those commits for this PR (down to one).

spohnan:
rebased to pick up recent merges and squashed into a single commit

chrisbennight:
Hm, that first test keeps failing - though I don' see hwo this pull request would have caused it.  The error though points to a verison mismatch in the gs-web-app 


```
jar:file:/home/travis/build/ngageoint/geowave/geowave-test/target/geoserver/WEB-INF/lib/gs-main-2.6.2.jar!/applicationContext.xml]: 
Cannot resolve reference to bean 'resourceLoader' while setting constructor argument; 
  nested exception is org.springframework.beans.factory.BeanCreationException: 
    Error creating bean with name 'resourceLoader' defined in URL   
      [jar:file:/home/travis/build/ngageoint/geowave/geowave-test/target/geoserver/WEB-INF/lib/gs-main-2.6.2.jar
```

2.6.2 is wrong for this test - should be 2.5.5

jwomeara:
Are we pulling a system property for the geoserver version directly in a
child pom?  That's the issue we had before and we fixed it by specifying
versions through dependency management in the parent pom. I am not by a
computer at the moment, but I can check tomorrow.

rfecher:
I see nothing in this commit that should cause travis issues on gs-web-app.  Let's see if other builds consistently have this issue too.  I have a build matrix running in travis right now.

jwomeara:
I left a note on the line of code that I believe is causing the problem.  It was pulled with merge request #281

jwomeara:
Actually, generally I see that in the test pom, the geoserver.version property is used in a bunch of places.  It shouldn't be as this defeats the purpose of dependency management.  By doing this I think that we are forcing the test pom to ignore the system property specified on the command line and instead it is using the default value specified in the parent pom.  If this is the real cause of the issue, then we just need to remove the version specifier from all of the geoserver dependencies...  I also think this is my goof as the commit log shows that it's been this way since I did the initial commit...

rfecher:
the strange thing is that this commit seems harmless

jwomeara:
Maybe the new geotools release affected the older jars in the boundless
repo...?

chrisbennight:
I deleted all travis caches in case there was something old in there; when I run the same build locally there's no 2.6x jar's in the lib directory.

spohnan:
will remove all pom and build script changes and resubmit so this truly just updates images and web pages. A second pull request will be used to isolate any possible problems with other changes.

PR #293, opened 2015-03-26
Documentation and web site updates

spohnan:
Reduce size of images used in docs and site
Fix RPM download link


chrisbennight:
ship it!

Issue #285, opened 2015-03-24
Exception on input format with large / complex multi-polygon


chrisbennight:
Exceptions in map reduce based query have been reported when using large multi-polygons (large as in large number of vertices)

chrisbennight:
So I was able to duplicate an error that may be related.  It's just a basic java memory limit - based probably on the environmental settings for the iterators - at some point the geometries appear to get big enough that they exhaust it  Stack trace at bottom:

This is happening for me at values over 29,000 vertices in a polygon. (with default memory settings, etc.)

Here's the test I used for this: https://github.com/ngageoint/geowave/commit/e34e4709ba1a8975f60f76310eafd5ee3d14c449

Not sure what the best solution is
  * Document it as a limit
  * Optional / integrated simplify capability
  * Tile the query geometry into multiple smaller queries (ideally based on vertex density?)

?



```
java.lang.OutOfMemoryError: GC overhead limit exceeded
	at com.vividsolutions.jts.geom.impl.CoordinateArraySequence.<init>(CoordinateArraySequence.java:116)
	at com.vividsolutions.jts.geom.impl.CoordinateArraySequenceFactory.create(CoordinateArraySequenceFactory.java:91)
	at com.vividsolutions.jts.io.WKBReader.readCoordinateSequence(WKBReader.java:310)
	at com.vividsolutions.jts.io.WKBReader.readCoordinateSequenceRing(WKBReader.java:333)
	at com.vividsolutions.jts.io.WKBReader.readLinearRing(WKBReader.java:241)
	at com.vividsolutions.jts.io.WKBReader.readPolygon(WKBReader.java:252)
	at com.vividsolutions.jts.io.WKBReader.readGeometry(WKBReader.java:191)
	at com.vividsolutions.jts.io.WKBReader.read(WKBReader.java:152)
	at com.vividsolutions.jts.io.WKBReader.read(WKBReader.java:133)
	at mil.nga.giat.geowave.store.GeometryUtils.geometryFromBinary(GeometryUtils.java:198)
	at mil.nga.giat.geowave.store.filter.SpatialQueryFilter.fromBinary(SpatialQueryFilter.java:191)
	at mil.nga.giat.geowave.index.PersistenceUtils.fromBinary(PersistenceUtils.java:92)
	at mil.nga.giat.geowave.store.filter.DistributableFilterList.fromBinary(DistributableFilterList.java:58)
[...]

PR #297, opened 2015-03-28
added large polygon query, geometry interning in filters

chrisbennight:
#285


geowave-accumulo/pom.xml
```diff
@@ -41,7 +41,7 @@
 				</exclusion>
 			</exclusions>
 		</dependency>
-		<dependency>
+        <dependency>
```
rfecher: this looks to me like it was previously formatted correctly - are you trying to pad your stats? ;)

geowave-store/src/main/java/mil/nga/giat/geowave/store/filter/SpatialQueryFilter.java
```diff
@@ -188,7 +193,7 @@ public void fromBinary(
 					fieldId));
 		}
 		buf.get(theRest);
-		queryGeometry = GeometryUtils.geometryFromBinary(geometryBinary);
+		queryGeometry = geometryInterner.intern(GeometryUtils.geometryFromBinary(geometryBinary));
```
rfecher: fancy! But in looking at JTS geometry I'm not convinced that it will work - I saw this in the source:
"#equals(Object) and #hashCode are not overridden, so that when two topologically equal Geometries are added to HashMaps and HashSets, they remain distinct. This behaviour is desired in many cases."

geowave-accumulo/pom.xml
chrisbennight: would you believe feng shui?

geowave-store/src/main/java/mil/nga/giat/geowave/store/filter/SpatialQueryFilter.java
chrisbennight: I should be able to add a test in there for this;  it's probably worth testing for regardless in case implementations change.  But it does have an effect:

Without it
https://travis-ci.org/ngageoint/geowave/builds/56052150

With it
https://travis-ci.org/ngageoint/geowave/builds/56170108

chrisbennight:
```Java
@Test
	public void testInterning()
			throws ParseException {
		final Geometry g = GeometryUtils.GEOMETRY_FACTORY.createPolygon(new Coordinate[] {new Coordinate(0,0), new Coordinate(1,0), new Coordinate(1,1), new Coordinate(0,1), new Coordinate(0,0)});
		final Geometry gNewInstance = GeometryUtils.GEOMETRY_FACTORY.createPolygon(new Coordinate[] {new Coordinate(0,0), new Coordinate(1,0), new Coordinate(1,1), new Coordinate(0,1), new Coordinate(0,0)});
		final WKBWriter wkbWriter = new WKBWriter();
		final byte[] b = wkbWriter.write(g);
		final byte[] b2 = new byte[b.length];
		System.arraycopy(b, 0, b2, 0, b.length);
		final WKBReader wkbReader = new WKBReader();
		final Geometry gSerialized = wkbReader.read(b);
		final Geometry gSerlizedArrayCopy = wkbReader.read(b2);


		System.out.println("Orig = new instance, same constructor: " + g.equals(gNewInstance));
		System.out.println("Orig  -> wkb -> geom equals: " + g.equals(gSerialized));
		System.out.println("Orig -> wkb -> arraycopy -> geom equals: " + g.equals(gSerlizedArrayCopy));
		System.out.println("(Orig -> wkb -> geom) = (wkb -> arraycopy -> geom): " + gSerialized.equals(gSerlizedArrayCopy));
	}
```

prints:
```
Orig = new instance, same constructor: true
Orig  -> wkb -> geom equals: true
Orig -> wkb -> arraycopy -> geom equals: true
(Orig -> wkb -> geom) = (wkb -> arraycopy -> geom): true
```

which seems to disagree with comment above, but explain why the interner worked; digging into it.

chrisbennight: Here you go, from the JTS source:

```Java
  /**
  * Tests whether this geometry is 
  * topologically equal to the argument geometry.
   * <p>
   * This method is included for backward compatibility reasons.
   * It has been superseded by the {@link #equalsTopo(Geometry)} method,
   * which has been named to clearly denote its functionality.
   * <p>
   * This method should NOT be confused with the method 
   * {@link #equals(Object)}, which implements 
   * an exact equality comparison.
   *
   *@param  g  the <code>Geometry</code> with which to compare this <code>Geometry</code>
   *@return true if the two <code>Geometry</code>s are topologically equal
   *
   *@see #equalsTopo(Geometry)
   */
  public boolean equals(Geometry g) {
    return equalsTopo(g);
  }
```

geowave-accumulo/pom.xml
chrisbennight: Challenge accepted

geowave-store/src/main/java/mil/nga/giat/geowave/store/filter/SpatialQueryFilter.java
rfecher: You win again, before I just googled "JTS Geometry source" to be sure equals()/hashCode() were overridden to do topological equality - it turns out I was looking at version 1.11 source...

It changed since then and you're correct we are using a version that overrides equals and hashcode:
http://sourceforge.net/p/jts-topo-suite/code/321/

PR #298, opened 2015-03-29
Findbugs; add to verify phase, address current issues

chrisbennight:
#228 
Also added findbugs to the build now;  had to exempt a few visitors  (see the omitVisitors property) - all of them basically caused by 
  * returning array types (byte[] value)
  * static non private instances of array types (or other objects)

it's that mutable state / security warning we talked about.  There's ~90 of those across the whole project that are not tracked.  Not necessarily issues, but probably need to take the time to verify before we annotate them directly.

But for any of the other errors findbugs will now fail the build.   The error will be printed in the mvn output, but the easiest way is either to get a findbugs plugin in your ide, or if you type

```
mvn findbugs:gui
```

a window detailing and explaining the issues will pop up for each module build

---------------------

also hits an admittedly unrelated javadoc issue - the -Xdoclint:none command needed to make java8 not blow up causes java7 to blow up.  This was preventing the install phase from finishing, which caused the verify phase to pull old libraries from the cache, which led to random results - so resolved it here (with a profile that injects the argument and only activates for java8+)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ cache:
   directories:
   - $HOME/.m2
 install: "mvn -q clean install javadoc:aggregate -Dfindbugs.skip -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -DskipITs=true -DskipTests=true -P ${PLATFORM_VERSION}; .utility/build-docs-site.sh"
-script: "mvn -q verify -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -P ${PLATFORM_VERSION}"
+script: "mvn -q -T 2C verify -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -P ${PLATFORM_VERSION}"
 before_install:
   - export MAVEN_OPTS="-Xmx512m -XX:MaxPermSize=192m"
   - chmod +x .utility/push-javadoc-to-gh-pages.sh
--- a/geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/util/EntryIteratorWrapper.java
+++ b/geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/util/EntryIteratorWrapper.java
@@ -94,6 +94,9 @@ public class EntryIteratorWrapper<T> implements
 	public T next()
 			throws NoSuchElementException {
 		final T previousNext = nextValue;
+		if (nextValue == null) {
+			throw new NoSuchElementException();
+		}
 		nextValue = null;
 		return previousNext;
 	}
--- a/geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/distance/CoordinateCircleDistanceFn.java
+++ b/geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/distance/CoordinateCircleDistanceFn.java
@@ -12,7 +12,7 @@ public class CoordinateCircleDistanceFn implements
 		DistanceFn<Coordinate>
 {
 
-	protected static CoordinateReferenceSystem DEFAULT_CRS;
+	protected static final CoordinateReferenceSystem DEFAULT_CRS;
 	static {
 		try {
 			DEFAULT_CRS = CRS.decode(
--- a/geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/kmeans/mapreduce/KMeansMapReduce.java
+++ b/geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/kmeans/mapreduce/KMeansMapReduce.java
@@ -1,8 +1,7 @@
 package mil.nga.giat.geowave.analytics.kmeans.mapreduce;
 
-import java.io.IOException;
-import java.util.List;
-
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.Point;
 import mil.nga.giat.geowave.accumulo.mapreduce.GeoWaveWritableInputMapper;
 import mil.nga.giat.geowave.accumulo.mapreduce.input.GeoWaveInputKey;
 import mil.nga.giat.geowave.accumulo.mapreduce.output.GeoWaveOutputKey;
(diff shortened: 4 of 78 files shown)


.travis.yml
```diff
@@ -17,7 +17,7 @@ cache:
   directories:
   - $HOME/.m2
 install: "mvn -q clean install javadoc:aggregate -Dfindbugs.skip -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -DskipITs=true -DskipTests=true -P ${PLATFORM_VERSION}; .utility/build-docs-site.sh"
-script: "mvn -q verify -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -P ${PLATFORM_VERSION}"
+script: "mvn -q -T 2C verify -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -P ${PLATFORM_VERSION}"
```
chrisbennight: 2 threads per core on builds;  seems to give a 2-10 minute speedup on travis.
Note that the install/javadoc version throws an exception using this.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/distance/CoordinateCircleDistanceFn.java
```diff
@@ -12,7 +12,7 @@
 		DistanceFn<Coordinate>
 {
 
-	protected static CoordinateReferenceSystem DEFAULT_CRS;
+	protected static final CoordinateReferenceSystem DEFAULT_CRS;
```
chrisbennight: now throws a runtime error in the static initializer if 4326 can't be loaded;  but really, what else can we do?  We are sort of dead in the water, so that's probably the best alternative?

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/kmeans/mapreduce/KSamplerMapReduce.java
```diff
@@ -242,7 +242,7 @@ protected void reduceNativeValues(
 			for (final Object value : values) {
 				final AnalyticItemWrapper<T> sampleItem = itemWrapperFactory.create((T) value);
 				Integer outputCount = outputCounts.get(groupID);
-				outputCount = outputCount == null ? 0 : outputCount;
+				outputCount = outputCount == null ? Integer.valueOf(0) : outputCount;
```
chrisbennight: Saves on unboxing

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/mapreduce/kde/compare/ComparisonAccumuloStatsReducer.java
```diff
@@ -109,7 +109,7 @@ protected void reduce(
 
 	private Point2d[] fromIndexToLL_UR(
 			final long index ) {
-		final double llLon = ((Math.floor(index / numYPosts) * 360.0) / numXPosts) - 180.0;
+		final double llLon = ((Math.floor(index / (double) numYPosts) * 360.0) / numXPosts) - 180.0;
```
chrisbennight: double check here, but since it's in a Math.floor it's being implicitly converter to a double regardless  (and for int/int there's no need to Math.floor()) - so I assume this is what was actually intended here.

geowave-benchmark/src/main/java/mil/nga/giat/geowave/benchmark/utils/HistogramGenerator.java
```diff
@@ -354,7 +354,6 @@ public void originalTest(
 		final List<Double> featsPerTile10 = new ArrayList<Double>();
 		final Map<Integer, Integer> collsPerTier10 = new HashMap<Integer, Integer>();
 		final Map<Integer, Integer> featsPerTier10 = new HashMap<Integer, Integer>();
-		final int[] hist10 = new int[500];
```
chrisbennight: @jwomeara 
A few deletions like this in HistogramGenerator - take a looks, it looks like the values were never actually used (they were written two, but never read.

geowave-index/pom.xml
```diff
@@ -23,7 +23,6 @@
         <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>annotations</artifactId>
-            <version>${findbugs.version}</version>
```
chrisbennight: apparently not rev'd concurrently;  findbugs is on 3.0.1, annotation on 3.0.0

geowave-raster/src/main/java/mil/nga/giat/geowave/raster/plugin/GeoWaveRasterReader.java
```diff
@@ -103,7 +103,7 @@
 
 	private static Set<AxisDirection> LEFTDirections;
 	// class initializer
-	{
+	static {
```
chrisbennight: get's rid of potential race condition writing from instance to static

geowave-test/src/main/java/mil/nga/giat/geowave/test/GeoWaveTestEnvironment.java
```diff
@@ -75,7 +77,8 @@
 			"./target/accumulo_temp"); // breaks on windows if temp directory
 										// isn't on same drive as project
 
-	protected static boolean DEFER_CLEANUP = false;
+	protected static final AtomicBoolean DEFER_CLEANUP = new AtomicBoolean(
```
chrisbennight: probably pointless based on how we access it, but possibly technically correct.

geowave-test/src/main/java/mil/nga/giat/geowave/test/service/ServicesTestEnvironment.java
```diff
@@ -96,8 +103,28 @@ public static void startServices() {
 					gsWebapp.setContextPath(GEOSERVER_CONTEXT_PATH);
 					gsWebapp.setWar(GEOSERVER_WAR_DIR);
 
-					final WebAppClassLoader classLoader = new WebAppClassLoader(
-							gsWebapp);
+					final WebAppClassLoader classLoader = AccessController.doPrivileged(new PrivilegedAction<WebAppClassLoader>() {
```
chrisbennight: Apparently all classloader instances should be instantiated inside a doPrivileged block.

geowave-vector/src/main/java/mil/nga/giat/geowave/vector/plugin/lock/AuthorizedLock.java
```diff
@@ -18,6 +19,9 @@
  * 
  * 
  */
+@SuppressFBWarnings({
+	"SE_TRANSIENT_FIELD_NOT_RESTORED"
```
chrisbennight: @rwgdrummer 
Comment call it out (after serialization/deserialization setlockmanager needs to be called), but is that enforced in any way;  (i.e. will geotools serialize this class somehow - and if so, do we ensure/test that we can cause setlockmanager to be called?

pom.xml
```diff
@@ -693,5 +695,14 @@
 				</repository>
 			</repositories>
 		</profile>
+		<profile>
+			<id>doclint-java8-disable</id>
```
chrisbennight: @spohnan This is what was breaking javadoc on jdk7 - the -Xdoclint:none  blows up there (it was just collapsed automatically by travis in the install phase block in the log).   This profile should only activate on jdk8+ - javadoc not looks for ${javadoc.opts}, so it's either empty (jdk7), or has the doclint string in it.

geowave-accumulo/src/main/java/mil/nga/giat/geowave/accumulo/util/EntryIteratorWrapper.java
```diff
@@ -94,6 +94,9 @@ public boolean hasNext() {
 	public T next()
 			throws NoSuchElementException {
 		final T previousNext = nextValue;
+		if (nextValue == null) {
```
chrisbennight: I think this logic is correct;  intent here (and in the other iterators changed in the same pattern) was to actually throw NoSuchElementException() when .next() was called on the end element.  (It looks like we would just return null in that case instead - but normally never got there because it was guarded with hasNext())

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/mapreduce/kde/compare/ComparisonAccumuloStatsReducer.java
rfecher: yeah, or I suppose we could just get rid of the floor.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/kmeans/mapreduce/KMeansMapReduce.java
```diff
@@ -76,59 +65,43 @@
...
-			@Override
-			public void notify(
+		AssociationNotification<Object> centroidAssociationFn = new AssociationNotification<Object>()
+		{
+			@Override public void notify(
 					final CentroidPairing<Object> pairing ) {
 				outputKeyWritable.set(
-						pairing.getCentroid().getGroupID(),
-						pairing.getCentroid().getID());
+						pairing.getCentroid().getGroupID(), pairing.getCentroid().getID());
```
rfecher: the formatting for the added lines looks strange and the format for the removed lines looks correct, is this actually the correct format?

geowave-test/src/main/java/mil/nga/giat/geowave/test/GeoWaveTestEnvironment.java
rfecher: I'm happy with the added safety, but am confident its not necessary given the usage of the mutex around any access to this flag.

geowave-vector/src/main/java/mil/nga/giat/geowave/vector/plugin/GeoWaveGTDataStore.java
```diff
@@ -112,6 +112,8 @@
 			LOGGER.error(
 					"Unable to decode EPSG:4326 CRS",
 					e);
+			throw new RuntimeException(
```
rfecher: pass along the factory exception?
ie. use constructor:
throw new RuntimeException(<String message>, e);

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/mapreduce/kde/compare/ComparisonAccumuloStatsReducer.java
chrisbennight: But, but, floor rounds to -infinity,  int/int rounds to zero!  What if index or numYPosts is negative! :)

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/kmeans/mapreduce/KMeansMapReduce.java
chrisbennight: I blame this on the eclipse formatter.  I downloaded luna, imported the format, and ran it on the file and it doesn't change anything.   If I make, say, pairing.getCentroid().getGroupID(), pairing.getCentroid().getID());  take two lines it doesn't change it back though either.

chrisbennight: ah hah.   "@formatter: on"  !=  "@formatter:on".    The extra space caused the formatter to never turn on.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/kryo/FeatureSerializer.java
```diff
@@ -19,6 +20,9 @@
 		Serializer<SimpleFeature>
 {
 
+	@SuppressFBWarnings({
+		"RR_NOT_CHECKED"
+	})
```
rwgdrummer: Yep.  Was working this a bit more last week.

PR #304, opened 2015-04-01
Added logic to test bounding box statistics

jwomeara:


jwomeara:
I'm trying to get this to work through Jace, so I updated our unit tests as a sanity check...

Issue #306, opened 2015-04-02
WFS-T.  BBOX parameter fails in tablet with GeoServer 2.6.x


rwgdrummer:
Example:
 curl "http://localhost:8080/geoserver/wfs?service=wfs&version=1.1.0&request=GetFeature&typename=cite:mt&bbox=-66.929286803243,-27.946435572149,42.839828269075,28.7418"

Reason for failure: geoserver does not autofill the geometry descriptor for the filter.

In addition, this ticket addresses:
(1) Serialization for Analytics to support Spark.  
(2) Removing a suppressed warning for findbugs in Kryo serialization and makign an appropriate fix.
(3) Starting to move towards removing accumulo specific dependencies in analytics

PR #307, opened 2015-04-02
GEOWAVE-306

rwgdrummer:
#306


geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/clustering/CentroidManagerGeoWave.java
```diff
@@ -283,12 +291,23 @@ public CentroidManagerGeoWave(
...
+					this.getClass(),
+					BasicAccumuloOperationsFactory.class,
+					DirectBasicAccumuloOperationsFactory.class).build(
+					runTimeProperties.getProperty(GlobalParameters.Global.ZOOKEEKER),
+					runTimeProperties.getProperty(GlobalParameters.Global.ACCUMULO_INSTANCE),
+					runTimeProperties.getProperty(GlobalParameters.Global.ACCUMULO_USER),
+					runTimeProperties.getProperty(GlobalParameters.Global.ACCUMULO_PASSWORD),
+					runTimeProperties.getProperty(GlobalParameters.Global.ACCUMULO_NAMESPACE));
+		}
+		catch (InstantiationException | IllegalAccessException e) {
```
chrisbennight: What's the intent behind converting these exceptions all to IOException?

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
```diff
@@ -68,46 +77,40 @@ public Object get(
...
-				LOGGER.info("Setting parameter : " + pName + " to " + value.toString());
-				properties.put(
-						pName,
-						value);
-				return value;
-			}
-
-			return properties.get(pName);
+		if (!properties.containsKey(pName)) {
+			LOGGER.info("Setting parameter : " + pName + " to " + value.toString());
```
chrisbennight: Instead of string concatenation, move the LOGGER statements over to use dynamic content injection {}
ref: http://slf4j.org/faq.html#logging_performance
String won't even be built if the log level is high enough.

jwomeara: Nice!  Didn't even know you could do that.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
```diff
@@ -68,46 +77,40 @@ public Object get(
...
-	public void store(
+	public synchronized void store(
 			final ParameterEnum propertyName,
 			final Object value ) {
-		synchronized (MUTEX) {
-			properties.put(
-					toPropertyName(propertyName),
-					value);
-		}
+		properties.put(
```
chrisbennight: Minor nit, but probably should be properties.setProperty()
(I think all it does is enforce (String, String), but a good habbit)

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
```diff
@@ -399,21 +419,19 @@ public DistributableQuery getPropertyAsQuery(
...
-				else {
-					properties.put(
-							option.getLongOpt(),
-							option.getValue());
-				}
+		for (final Option option : commandLine.getOptions()) {
+			if (!option.hasArg()) {
+				properties.put(
+						option.getLongOpt(),
+						true);
```
chrisbennight: Not sure what will happen here - not supposto have anything but string, string in a properties, but this snuck in using the put method.  Might implicitly get converted, but probably better to be explicit.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/clustering/CentroidManagerGeoWave.java
rwgdrummer: Used in situations such as setup in Mapper/Reducers that only handles IOException.  Rather littering the method signature and pushing the redundant handling else where, I do it here.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
rwgdrummer: True,  Can do this.  I know SLF4j supports this.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
rwgdrummer: Then I will change it to a map.  The values need to serializable objects.  I will make the adjustment.

rwgdrummer:
I started looking at this again more closely.
PropertyManagement changed a bit for me as I started seeing its potential as a uniform way to handle parameters in the analytics. Looking at it some more tonight, I think there is some thought that I left unfinished.  I am going to make some additional adjustments tomorrow.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/clustering/CentroidManagerGeoWave.java
chrisbennight: Gotcha - I understand it in the case of a constrained method that needs to be implemented/overrided - just didn't see that here.   What does e.getMessage() do in this case?   Is there any value in something like  
```Java
throw new IOException(e, e.getMessage())
```
(might be pointless - intent here is just to make sure we aren't slightly obfuscating the InstatiantionException of IllegalAcessException, since those are more useful than IOException)

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
chrisbennight: That works to!

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
chrisbennight: And to be fair we still have to move over the rest of the codebase to this as well.

rfecher: it sounds like someone is going to be padding his github stats this weekend ;)

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/tools/PropertyManagement.java
rwgdrummer: Thanks.

geowave-analytics/src/main/java/mil/nga/giat/geowave/analytics/clustering/CentroidManagerGeoWave.java
rwgdrummer: yep.

rwgdrummer:
Addressed the concerns of the last push.
(1) Added localized message to IOException
(2) Consistent handling of serializable objects in the parameter management.
(3) formatting efficiencies.
So, ready for another review.
Thanks

PR #308, opened 2015-04-03
Added a fix to convert to default CRS before storing bounding box stats.

jwomeara:


jwomeara:
Somebody want to check this out for me?

I updated the code to automatically convert entries to our default CRS when recording bounding box stats for feature data.  I added a test for this as well.

I also made a utility class for the reprojection method that we have been using in a few places.  There are also a few random 'code cleanup' and 'formatter' changes.

rwgdrummer:
+1.  
Looks good.  Will wait till travis finish.

geowave-vector/src/main/java/mil/nga/giat/geowave/vector/util/FeatureDataUtils.java
```diff
@@ -0,0 +1,79 @@
...
+				// this feature differs from the persisted type in CRS,
+				// calculate the transform
+				try {
+					featureTransform = CRS.findMathTransform(
+							crs,
+							GeoWaveGTDataStore.DEFAULT_CRS,
+							true);
+				}
+				catch (final FactoryException e) {
+					LOGGER.warn(
```
chrisbennight: I know you were re-factoring existing code, but in the case of not being able to transform to EPSG:4326 do we really want to return an feature in the incorrect crs (presumably to be uploaded)?  (Caveat:  I didn't check upstream of this code to see if there's a CRS check)

I would think it would be better not to upload and throw an error in the log (or maybe even runtime exception, though leaning away from that).   Probably last thing we want is to upload inconsistent info.

Or I could be crazy, thoughts?

jwomeara: The comments explain the original thought process a bit. It seems
reasonable to me though to log an error and cancel the operation when we
can't convert to the default CRS.

Want me to make a ticket or update the existing pull request?

chrisbennight: Comment only helps if people read it.  Whoops.   
New ticket is fine - I can make one

PR #314, opened 2015-04-07
Roll to v0.8.5-SNAPSHOT

spohnan:


drew-bower:
@chrisbennight Does this snapshot include the large polygon optimization?

@panrufio, @djohnson729

chrisbennight:
@drew-bower @panrufio @djohnson729

The 0.8.4 Tag has it (so does this).  Need to add it to the main page, but here's our changelog:
http://ngageoint.github.io/geowave/changelog.html

The item is #285 - Exception on input format with large / complex multi-polygon

Issue #318, opened 2015-04-09
Clean up readme


chrisbennight:
Since we have better documentation now we can remove outdated info from the readme, referr to the readme, and just cover the high level details.

PR #319, opened 2015-04-09
Cleaned up README

chrisbennight:


chrisbennight:
@jwomeara 
can you double check the PDAL and Mapnik versions I listed here

chrisbennight:
#318

Issue #323, opened 2015-04-10
Vagrant config should do shallow git clone


chrisbennight:
don't need the entire history;  doing a shallow clone (--depth 10 or so) should be fine

PR #324, opened 2015-04-10
Reduce vagrant geowave clone depth

chrisbennight:


chrisbennight:
#323

chrisbennight:
@jwomeara

jwomeara:
The change is fine with me.  It all depends on how we expect the vagrant setup to be used.  If it's just for checking things out, a minimal history is fine.  However, if we want people to use it as a dev environment, we might want the full history.  

I think at this point the jdk download is the most time consuming task.

chrisbennight:
do you need the full history to develop?  I think you can still do a branch and work like normal, no?  I'll test it real quick

jwomeara:
If that's the case, then merge away.

chrisbennight:
https://github.com/ngageoint/geowave/tree/shallow_clone_test

PR #325, opened 2015-04-10
Added accumulo exceptions to Jace Proxies

jwomeara:


jwomeara:
This is a flat file which won't affect the build.  Looks good to me!

Issue #263, opened 2015-03-14
Coveralls code coverage is incorrectly low due to lack of merge between unit and integration tests


chrisbennight:
ref:
https://github.com/trautonen/coveralls-maven-plugin/issues/64#issuecomment-73431035

Looks like whichever runs last (it, ut) for a particular class "wins".
Seems like we can move over to the merge goal of jacoco and run coverage against that.

chrisbennight:
After battling this for a bit, I'm voting we just host a move our coverage to a Sonar instance;  most of the online tools barely support java (php, python, js, ruby focused mostly), and those that do (or have user contributed methods to, like coveralls), have poor multi-module support (resolving back to either stateful report merging, or jacoco multi-module issues).

chrisbennight:
Fixed by @rfecher with a huge ant task that merges everything together.

PR #329, opened 2015-04-10
fixed coveralls to report both integration and unit tests (#263)

rfecher:


chrisbennight:
no mas. won't work as is because the 
```mvn coveralls:report```
command fails on the parent pom with
```[ERROR] No plugin found for prefix 'coveralls' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/travis/.m2/repository), central (https://repo.maven.apache.org/maven2)] -> [Help 1]```

I tried tweaking the .travis.yml to run the coveralls command in the geowave-test directory, but it then complains that it can't find source for other files.

PR #337, opened 2015-04-12
Updated dependencies

chrisbennight:


rfecher:
I see you updated the test matrix to version n and n-1 for geotools/geoserver but wouldn't we want to also update the parent pom to be version n (13.0 and 2.7.0)

chrisbennight:
Yes we would!

rfecher:
done...just waiting for travis now

PR #341, opened 2015-04-15
made CLI operations pluggable (#340)

rfecher:


chrisbennight:
Build failed because of findbugs "dubious code use" - it seems you thought you could initialize a new PrintWriter without specifying a default charset.  DENIED!

There was also an issue with the antrun task in geowave-test;  if you did a clean checkout AND skipped tests/verify phases then the target/site directory would never get created, and the antrun task would fail.  I just added a task to pre-create that directory.

rfecher:
admittedly I ran it locally with findbugs and it failed, so I skipped findbugs...

Issue #343, opened 2015-04-16
FID Filter being ignored in GeoServer plug-in


rwgdrummer:
FId Filter is an object returned by CQLtoFilter translations when a FID is directly referenced in a filter.  This allows the plug-in to issue a query to the alternate index.  Changes in a prior ticket caused the optimization code to be bypassed.

PR #345, opened 2015-04-16
Geowave 343

rwgdrummer:
#343


rfecher:
looks reasonable to me, waiting for travis...

PR #347, opened 2015-04-17
Update Jenkins URL to use hostname instead of IP

spohnan:
I had to recreate the jenkins server in AWS to have it pick up an IAM role so the IP changed. This change will have it use hostname until such time as this setting goes away and we just start using public S3 buckets instead of jenkins for most of our artifacts.


spohnan:
Self accepting this request as it's just a single line config change of our jenkins server.

PR #354, opened 2015-04-28
GEOWAVE-352

rwgdrummer:
#352 

Command line tool adjusted to take -r (for recipient) argument to allocate transaction ids for the specific user.


chrisbennight:
Looks good!

Issue #320, opened 2015-04-09
Refactor projects to separate core and plugins


rfecher:
Here's an idea of how it could be laid out (some new projects added to give an idea of how the structure may grow).

    ├───.utility
    ├───analytics
    │      ├───api
    │      ├───mapreduce
    │      ├───spark
    ├───core
    │      ├───cli
    │      ├───geotime
    │      ├───index
    │      ├───ingest
    │      └───store
    ├───deploy
    ├───dev-resources
    ├───docs
    ├───examples
    ├───extensions
    │      ├───datastores
    │      │     ├───accumulo
    │      │     ├───cassandra
    │      │     ├───hbase
    │      │     └───simplehdfs
    │      └───formats
    │           ├───geolife
    │           ├───geotools-raster
    │           ├───geotools-vector
    │           ├───gpx
    │           ├───stanag4607
    │           ├───stanag4676
    │           └───tdrive
    ├───services
    │      ├───api
    │      ├───client
    │      │      ├───java
    │      │      ├───javascript
    │      │      └───ruby
    │      └───webapp
    └───test

chrisbennight:
I think having a lightweight tools (or something like it) module that just has items that need to be referenced in multiple modules (in project and inherited) is a good idea.   I'm thinking the eclipse formatter, and I'd like to do a findbugs configuration file, and a checkstyle configuration file.

With them being owned by an artifact it's easy to reference them from either the parent pom directory, or from within a child pom.   If you don't do this then paths are terminally broken (you can either configure for invocation from parent directory, or module subdirectory, but not both).

See: http://mblanchette.github.io/maven-java-formatter-plugin/0.4/examples.html#Multimodule_Configuration for example

(i.e. I'd be tempted to move benchmark out of tools.   What would you think of making a geowave-benchmark project?  I don't think the benchmark needs to be in the core geowave project.)

viggyprabhu:
Any particular reason why all the different projects are not maintained as a separate project code wise. The complete project can be built separately once in a while and released as Tarballs with all necessary jars for users but other wise, we can maintain projects separately. This will ensure very clear dependency separation and hierarchy. In such cases, atleast all the Plugin and tools projects can be maintained separately.

rfecher:
[...]

PR #360, opened 2015-04-30
Refactor project to separate core from extensions (#320)

rfecher:
Generally speaking there is now a hierarchy of projects to allow the project to more easily grow, rather than the previous flat module structure.

This did introduce breaking changes to the persistence model.

Here's a set of points covered within this issue:
* To separate Spatial temporal core implementation details from the more general purpose multi-dimensional index and key-value store, a new core module, 'geotime' was created.  To separate all spatial-temporal components from the core, field readers and writers are now pluggable through SPI.

* Also, analytics and ingest are now both able to run through GeoWaveMain using SPI to inject commandline operations (and there's a new "cli" module that contains the core for the command line tooling).

* The ingest framework now allows the optional dimensionality types to be injected through SPI, in which the geotime module injects "spatial" and "spatial-temporal".

* The different ingest types within geowave-types are now broken out individually in extensions/formats (and the term "type" is replaced by the term "format" to avoid confusion between ingest formats and data adapters).

* Analytics was quickly broken out between "api" and "mapreduce" but perhaps its worth revisiting to look more closely at the logical breakdown, because I didn't want to go down too deep a rabbit hole with this initial cut.

* Services were broken out between the service API (jax-rs interfaces), the java client stub (and if we were to provide more languages as clients than just java, this could become a directory), and the service implementation webapp

* "packaging" was placed in deploy to limit the top-level directories

* the eclipse formatter and cleanup were place in dev-resources as a standalone maven module so projects do not depend on a relative path to the formatter when running the maven format goal - therefore in order to run maven commands for geowave, the dev-resources must be installed first.  We should look to publish this artifact to maven central.

* deleted benchmarks and vagrant setup from the root geowave project and separated each into its own project on github/ngageoint


rwgdrummer:
Should we add a new category for Analytics rather than using the ingest category (CLI)?

.travis.yml
```diff
@@ -25,10 +25,11 @@ before_install:
   - chmod +x .utility/maven-coveralls-hack.sh
   - chmod +x .utility/changelog-generator.sh
   - .utility/maven-coveralls-hack.sh
+  - cd dev-resources;mvn clean install;cd ..
 after_success:
   - .utility/changelog-generator.sh
   - .utility/push-javadoc-to-gh-pages.sh
-  - cd geowave-test; mvn coveralls:report -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -P ${PLATFORM_VERSION}
+  - mvn coveralls:report -Daccumulo.version=${ACCUMULO_VERSION} ${ACCUMULO_LEGACY} -Dhadoop.version=${HADOOP_VERSION} -Dgeotools.version=${GEOTOOLS_VERSION} -Dgeoserver.version=${GEOSERVER_VERSION} -P ${PLATFORM_VERSION}
```
chrisbennight: probably should be "cd test;" at the beginning based on the coveralls config

rfecher: thanks, added 'cd test;'

also, there was an issue with jacoco reporting on both the munged and unmunged class directories, so I specifically excluded unmunged directories for the modules that use munge

rfecher:
good catch on the CLI analytic category, the category class existed, just the provider was still using the ingest category - it should be fixed now

chrisbennight:
* Probably small updates to #320 needed (vagrant, benchmark moving outside).
  * Should add a documents page that gives an overview of the project layout (but that can be a separate ticket)
  * Think there are still issues with the jacoco/coveralls stuff;  looks like all the source directories are removed in the test pom - which I don't think will work (unless there's some other trick I'm not seeing)

Issue #157, opened 2014-12-19
implement geowavefileoutput format for high-speed ingest


rfecher:
analogue to accumulofileoutputformat

keith-turner:
Did something similar to this w/ [Fluo][1].  Fluo has a specialized data format that it stores in Accumulo like Geowave.   [FluoFileOutputFormat][2] leverages/wraps code from AccumuloFileOutputFormat.  Later while looking into wrapping AccumuloOutputFormat, I decided it would be easier to write something that simply produces mutations instead.   [MutationBuilder][3] does this and was much easier to write than something that wraps AccumuloOutputFormat.  I have thought about scrapping FluoFileOutputFormat and instead writing a simple class that produces Accumulo Keys.  A user could use this class to produces Keys that are written using AccumuloFileOutputFormat.   

[1]: https://github.com/fluo-io/fluo
[2]: https://github.com/fluo-io/fluo/blob/fc80d097c6df4b9e26ebd0c2bef38d8e2dc447c6/modules/mapreduce/src/main/java/io/fluo/mapreduce/FluoFileOutputFormat.java
[3]: https://github.com/fluo-io/fluo/blob/fc80d097c6df4b9e26ebd0c2bef38d8e2dc447c6/modules/mapreduce/src/main/java/io/fluo/mapreduce/MutationBuilder.java

rfecher:
Thanks Keith,  wrapping the AccumuloFileOutputFormat's record writer is probably the simplest approach for us to follow too.

dcy2003:
I initially started down the path of implementing a `GeowaveFileOutputFormat` (that wraps `AccumuloFileOutputFormat`) as specified in the details of this ticket.  After banging my head against this approach for a few days, I am encountering some of the same challenges that I believe Keith was alluding to, and I think the approach he is recommending makes more sense.  

Here is my attempt to shed a little more light on the issue and propose an alternative approach:

Per the documentation, when leveraging `AccumuloFileOutputFormat` to write output in the Accumulo data file format, 
> “care should be taken to write only sorted data (sorted by org.apache.accumulo.core.data.Key), as this is an important requirement of Accumulo data files.”

[...]

PR #367, opened 2015-05-05
GEOWAVE-157

dcy2003:
Implementation for #157 ...

The crux of the solution is within `AccumuloKeyValuePairGenerator` which produces Accumulo [Key,Value] pairs (Geowave-formatted) given an adapter and an index, similar to the way we currently produce Mutations.

To fully demonstrate the "flow" of this solution, I created an example MapReduce job that:
1. reads in some sample Geonames data (downloaded a bunch of points in Barbados)
2. constructs SimpleFeatures for the sample data
3. leverages AccumuloKeyValuePairGenerator to generate [Key,Value] pairs
4. writes the data to file using AccumuloFileOutputFormat (after MapReduce sorts the [Key,Value] pairs) 

Feedback welcome!

### DIFF
--- a/examples/src/main/java/mil/nga/giat/geowave/examples/ingest/bulk/GeonamesDataFileInputFormat.java
+++ b/examples/src/main/java/mil/nga/giat/geowave/examples/ingest/bulk/GeonamesDataFileInputFormat.java
@@ -1,0 +1,31 @@
+package mil.nga.giat.geowave.examples.ingest.bulk;
+
+import java.io.IOException;
+
+import org.apache.hadoop.io.LongWritable;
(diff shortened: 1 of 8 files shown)


rfecher:
I looked through the files, and it looks like a good example. If we choose to follow-on with this to add more re-usable components, one thought would be to try to minimize the constants defined near the top of SimpleFeatureToAccumuloKeyValueMapper to provide a re-usable mapper that can be given the index and adapter through the job context, and the key generator could be more hidden so that the implementor could call something like writeNativeObject and the key generator would then take the native object and write the collection of key value pairs to the context...anyways, just a thought, not sure if its worth putting more into this yet.

Issue #116, opened 2014-11-20
Additional geowave-examples


chrisbennight:
Create additional example case for:

  * Query (really just a geotools query, but should have an example, plus CQL info)
  * Utilizing geowave-index or geowave-store as a secondary index on another system

Issue #361, opened 2015-04-30
KMeansIT test has a seemingly random failure


rfecher:
I copied the following from the travis-ci log in a run where all that was changed was the pom version, 0.8.6 to 0.8.7-SNAPSHOT and the other 3 passed, this one failed (seems like it occurred randomly):

ACCUMULO_VERSION='1.5.2'
ACCUMULO_LEGACY='-Daccumulo.legacy'
HADOOP_VERSION='2.6.0'
GEOTOOLS_VERSION='12.2'
GEOSERVER_VERSION='2.6.2'
PLATFORM_VERSION='""'

testIngestAndQueryGeneralGpx(mil.nga.giat.geowave.test.mapreduce.GeoWaveKMeansIT)  Time elapsed: 184.644 sec  <<< FAILURE!
java.lang.AssertionError: d34124b0-37b7-44ec-b30a-1b9d380ceb96
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at mil.nga.giat.geowave.test.mapreduce.GeoWaveKMeansIT.countResults(GeoWaveKMeansIT.java:313)
	at mil.nga.giat.geowave.test.mapreduce.GeoWaveKMeansIT.runKPlusPlus(GeoWaveKMeansIT.java:196)
	at mil.nga.giat.geowave.test.mapreduce.GeoWaveKMeansIT.testIngestAndQueryGeneralGpx(GeoWaveKMeansIT.java:137)
Results :
Failed tests: 
  GeoWaveKMeansIT.testIngestAndQueryGeneralGpx:137->runKPlusPlus:196->countResults:313 d34124b0-37b7-44ec-b30a-1b9d380ceb96

PR #378, opened 2015-05-07
Added GeoTools query example

dcy2003:
GeoTools query example for #116

### DIFF
--- a/examples/src/main/java/mil/nga/giat/geowave/examples/query/GeotoolsQueryExample.java
+++ b/examples/src/main/java/mil/nga/giat/geowave/examples/query/GeotoolsQueryExample.java
@@ -1,0 +1,257 @@
+package mil.nga.giat.geowave.examples.query;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import mil.nga.giat.geowave.adapter.vector.FeatureDataAdapter;
+import mil.nga.giat.geowave.core.geotime.GeometryUtils;
+import mil.nga.giat.geowave.core.geotime.IndexType;
+import mil.nga.giat.geowave.core.geotime.store.query.SpatialQuery;
+import mil.nga.giat.geowave.core.store.CloseableIterator;
+import mil.nga.giat.geowave.core.store.DataStore;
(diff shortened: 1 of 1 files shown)


chrisbennight:
Didn't dig into the code, but from looking at the errors it looks like you might need to add a maven munge directive here - seems there might by a method being called that's not valid for accumulo 1.5

rfecher:
I didn't see it that way, at least in the last failure what I saw was it failing on a random zookeeper transaction related test unrelated to this example code.  I think there's at least two random failures in our tests, I ticketed one #361 and I probably should have added an issue for this before restarting.  We'll probably see it again.

chrisbennight:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:compile (default-compile) on project geowave-example: Compilation failure
[ERROR] /home/travis/build/ngageoint/geowave/examples/src/main/java/mil/nga/giat/geowave/examples/query/GeotoolsQueryExample.java:[133,14] error: cannot find symbol
```

line 133 is 
https://github.com/ngageoint/geowave/blob/GEOWAVE-116_1/examples/src/main/java/mil/nga/giat/geowave/examples/query/GeotoolsQueryExample.java#L133

or 
```
accumulo.getConnector(	ACCUMULO_USER, ACCUMULO_PASSWORD)
```
where accumulo is org.apache.accumulo.minicluster.MiniAccumuloCluster

the getConnector method is marked as "since 1.6.0"
https://github.com/apache/accumulo/blob/master/minicluster/src/main/java/org/apache/accumulo/minicluster/MiniAccumuloCluster.java#L110

.


I think I win! :)

jwomeara:
We test against 1.5.2 in travis.  :)

https://github.com/ngageoint/geowave/blob/master/.travis.yml#L7

Munge it up, brother!

chrisbennight:
To be fair, could possibly skip munge and just go with a 1.5 compatible method if it wasn't deprecated.

Hopefully when 1.7 get's out we can cute some of hte 1.5 support.

rfecher:
@chrisbennight you win this time ;)

I don't know what I was looking at with the transaction stuff - I'll just stick to the claim that the one test I had restarted happened to have the random issue before it got to compiling the examples.

dcy2003:
Sorry for the oversight - should be fixed now.  I took @chrisbennight recommended approach and simply avoided the use of a 1.6-only method.

PR #388, opened 2015-05-14
stats calculation tool using CLI framework (#387)

rfecher:


chrisbennight:
Looks good, linking issue: #387

PR #390, opened 2015-05-14
Geowave 385

rwgdrummer:


rwgdrummer:
FYI.  Walked through this with Rich.  Had an extra 'compensation' logic for adapting different versions of feature data adapter stored images.  Took the adaptive component out by left the version byte in there for future use.

Just to recap:
CRS:84 does not have an SRID...so it fails in several unit tests.
urn:x-ogc:def:crs:EPSG:4326 forces long/lat, as does the current method of passing 'true' to the CRS.decode method with code=EPSG:4326. 
EPSG:4326, in itself, cannot force lat/long.  It all depends on the authority within the system.  For us our purposes, this is ok.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/util/FeatureDataUtils.java
```diff
@@ -76,4 +82,56 @@ else if (crs != null) {
 		}
 		return defaultCRSEntry;
 	}
+
+	public static String getAxis(
+			final CoordinateReferenceSystem crs ) {
+		final CoordinateSystem cs = crs == null ? null : crs.getCoordinateSystem();
+		if (cs != null && cs.getDimension() > 0) return cs.getAxis(
+				0).getDirection().name().toString();
+		return "";
```
chrisbennight: do we really want to return an empty string if we can't get a CoordinateSystem?   
If we don't want to deal with error propagation back for a nonsense case them maybe use a guava precondition instead?

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/FeatureDataAdapter.java
```diff
@@ -369,15 +382,16 @@ protected Object defaultTypeDataFromBinary(
 					ex);
 		}
 
-		final byte[] encodedTypeBytes = new byte[bytes.length - typeNameBytes.length - namespaceBytes.length - fieldVisibilityAtributeNameBytes.length - visibilityManagementClassNameBytes.length - timeAndRangeBytes.length - 20];
+		final byte[] encodedTypeBytes = new byte[bytes.length - axisBytes.length - typeNameBytes.length - namespaceBytes.length - fieldVisibilityAtributeNameBytes.length - visibilityManagementClassNameBytes.length - timeAndRangeBytes.length - 25];
```
chrisbennight: thoughts on calling out the "25" value as a constant, or at least variable (to document what it's from - though it's relatively obvious here)?

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/util/FeatureDataUtils.java
rwgdrummer: Do to the complexity of some of the CRS managed in geotools, I am not sure that we guarantee that the absence of an axis is an erroneous condition.  I interpret such cases as undefined or assumed.  Although I could test a hand full of CRS.  If they all define an axis, then we could go the route of the error.  Or we could just accept the undefined case.   Then, the question is whether an formal exception is necessary in these cases, a defined return value or , in this case, an innocuous return value.

Thoughts?

rwgdrummer: Thinking it over a bit and having a quick discussion with Rich, decided to stick with our default, EPSG:4326 with Axis EAST.

PR #400, opened 2015-05-20
GEOWAVE-398

rwgdrummer:


chrisbennight:
Once comment totally unrelated to this request - might be a good idea to give these a more descriptive name now that we are automatically generating a changelog
http://ngageoint.github.io/geowave/changelog.html

since the title is all that will show up.

PR #403, opened 2015-05-21
GEOWAVE-398 & GEOWAVE-401.  DWITHIN corrected to handle the unit conv…

rwgdrummer:
…ersion properly from distance (meters, km, etc.) to decimal degrees (EPSG:4326 long/lat).  There is still a degree of error difficult to compensate for when dealing polygons (e.g a country).  Future work will adopt a more accurate approach consistent with other LocationTech projects

### DIFF
--- a/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
+++ b/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
@@ -1,0 +1,518 @@
+package mil.nga.giat.geowave.adapter.vector.utils;
+
+import java.awt.geom.Point2D;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+
+import javax.measure.unit.SI;
+import javax.measure.unit.Unit;
+
+import mil.nga.giat.geowave.adapter.vector.plugin.GeoWaveGTDataStore;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.log4j.Logger;
+import org.geotools.ows.bindings.UnitBinding;
+import org.geotools.referencing.GeodeticCalculator;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.cs.CoordinateSystem;
+import org.opengis.referencing.cs.CoordinateSystemAxis;
+import org.opengis.referencing.operation.TransformException;
+
+import com.google.uzaygezen.core.BitSetMath;
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.Envelope;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.GeometryCollection;
+import com.vividsolutions.jts.geom.GeometryFactory;
+import com.vividsolutions.jts.geom.MultiPolygon;
+import com.vividsolutions.jts.geom.Point;
+import com.vividsolutions.jts.geom.Polygon;
+
+public class GeometryUtils
+{
+
+	private static Logger LOGGER = Logger.getLogger(GeometryUtils.class);
+
+	/**
+	 * Build a buffer around a geometry
+	 * 
+	 * @param crs
+	 * @param geometry
+	 * @param distanceUnits
+	 * @param distance
+	 * @return
+	 * @throws TransformException
+	 */
+	public static final Pair<Geometry, Double> buffer(
+			final CoordinateReferenceSystem crs,
+			final Geometry geometry,
+			final String distanceUnits,
+			final double distance )
+			throws TransformException {
+		Unit<?> unit;
+		try {
+			unit = (Unit<?>) new UnitBinding().parse(
+					null,
+					distanceUnits);
+		}
+		catch (final Exception e) {
+			unit = SI.METER;
+			LOGGER.warn(
+					"Cannot lookup unit of measure " + distanceUnits,
+					e);
+		}
+		final double meterDistance = unit.getConverterTo(
+				SI.METER).convert(
+				distance);
+		final double degrees = distanceToDegrees(
+				crs,
+				geometry,
+				meterDistance);
+		// buffer does not respect the CRS; it uses simple cartesian math.
+		// nor does buffer handle dateline boundaries
+		return Pair.of(
+				adjustGeo(
+						crs,
+						geometry.buffer(degrees)),
+				degrees);
+
+	}
+
+	/**
+	 * Consume a geometry that may be over the ranges of the CRS (e.g date-line
+	 * crossing). Adjust for crossings with a multi-polygon instance where each
+	 * contained polygon represents a portion of the provided geometry longitude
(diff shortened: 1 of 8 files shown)


extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
```diff
@@ -0,0 +1,114 @@
...
+	/** farther point in longitudinal axis given a latitude */
+
+	private static Point farthestPoint(
+			Point point,
+			double meters ) {
+		final GeodeticCalculator calc = new GeodeticCalculator();
+		calc.setStartingGeographicPoint(
+				point.getX(),
+				point.getY());
+		calc.setDirection(
```
rfecher: consider handling dateline crossing here

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
```diff
@@ -0,0 +1,114 @@
...
+				geometry.getEnvelopeInternal(),
+				factory == null ? new GeometryFactory() : factory,
+				meters);
+	}
+
+	private static double distanceToDegrees(
+			Envelope env,
+			GeometryFactory factory,
+			double meters ) {
+		return Collections.max(Arrays.asList(
```
rfecher: I'd suggest we just pick a single point, either the centroid or one with the y value with maximum absolute value (nearest the poles).  I think by using a point with the y value nearest a pole you will get the max degree distance from meters so it should be the most inclusive if that is desired (I could see arguments for either case).

rwgdrummer:
Addressed one comment and left the other.  Thinking better to have a false positive than a negative.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
```diff
@@ -0,0 +1,236 @@
...
+				crs);
+		calc.setStartingGeographicPoint(
+				point.getX(),
+				point.getY());
+		calc.setDirection(
+				90,
+				meters);
+		Point2D dest2D = calc.getDestinationGeographicPoint();
+		// if this flips over the date line then try the other direction
+		if (sign(dest2D.getX()) != sign(point.getX())) {
```
rfecher: I think this really should cover us as much as we can for the dateline.  In the last commit, this check was really all I was alluding to.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
```diff
@@ -0,0 +1,236 @@
...
+		final double meterDistance = unit.getConverterTo(
+				SI.METER).convert(
+				distance);
+		final double degrees = distanceToDegrees(
+				crs,
+				geometry,
+				meterDistance);
+		// buffer does not respect the CRS; it uses simple cartesian math.
+		return Pair.of(
+				adjustGeo(
```
rfecher: I'd forgo this call to adjustGeo()

I don't think it quite does what we want it to do - the way I read it, the geometry will self-correct for dateline crossing by converting values outside of the range of valid longitude values to be the reflection of that longitude value (ie. converting -181 to 179 for example).

The problem is that JTS is working in 2D cartesian space and this won't really result in a geometry that represents  what you want it to.  In fact I think it may often result in exceptions from JTS because the geometry will be invalid (folding across itself).  What is really need to correctly represent the geometry is to cut it into multiple geometries, on each side of the dateline (in some cases more than one in fact, if you think of a zigzag, 'Z' shape as an example).  I've written some code to basically do that, I think its out of scope for this right now though and I'd just get rid of adjustGeo().

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
```diff
@@ -0,0 +1,236 @@
...
+		final CoordinateSystem coordinateSystem = crs.getCoordinateSystem();
+		if (coordinateSystem.getDimension() > dimension) {
+			double lowerBound = coordinateSystem.getAxis(
+					dimension).getMinimumValue();
+			double bound = coordinateSystem.getAxis(
+					dimension).getMaximumValue() - lowerBound;
+			double sign = sign(val);
+			// re-scale to 0 to n, then determine how many times to 'loop
+			// around'
+			double mult = Math.floor(Math.abs((val + sign * (-1.0 * lowerBound)) / bound));
```
rfecher: is this assuming the absolute value of lowerBound is essentially upperBound?  Works for longitude so I don't care and in any case I don't think it really matters, I'd just suggest forgoing adjustGeo().

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
rwgdrummer: If I do not keep this adjustment code, then the geometries produced by buffer on invalid. This code is focused fixing the coordinate space issue. Thus, I think this needs to stay.  

That being said, the other issue is the wrapping problem you are referring to.  The problem will likely arise when performing the intersection (after the the DWITHIN conversion).
I agree that cutting into a multi-polygon is  the best option. My thought is to have another function take the wrapped geometry and convert, using that method when converting DWITHIN to intersection.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
rwgdrummer: Right.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/utils/GeometryUtils.java
```diff
@@ -0,0 +1,515 @@
...
+
+	/**
+	 * Make sure the coordinate falls in the range of provided coordinate
+	 * reference systems's coordinate system.
+	 * 
+	 * @param crs
+	 * @param coord
+	 * @return
+	 */
+	public static Coordinate adjustCoordinateToFitInRange(
```
dcy2003: This method does not always give proper results.  For example, it currently transforms long=116, lat=400 to long=116, lat=90 when the proper result should be long=116, lat=40.

Why do we call adjustCoordinateDimensionToRange() for the X coordinate but clipRange() for the Y and Z coordinates?  Shouldn't all three coordinates rely on adjustCoordinateDimensionToRange()?  It looks like the issues are stemming from clipRange()...

rfecher: adjustCoordinateDimensionToRange() is just the wrong math to do with y and z - they don't have the same reflection property at the extrema.

90 + 1 degrees latitude is arguably 89 degrees instead of clamping at 90 but its definitely not -89 (you wouldn't want to shift from north to south pole)

PR #404, opened 2015-05-22
Build dependencies updated for geoserver 2.7.1

spohnan:
http://blog.geoserver.org/2015/05/20/geoserver-2-7-1-released/


rfecher:
I went ahead and changed the parent pom to reflect the updated versions as defaults and just squashed it into your commit.

PR #420, opened 2015-06-05
Source config file if present for GeoServer setttings

spohnan:
This will allow us to add GeoServer configs like Xmx2g on an app server node in an external file either manually or with something like Puppet. This will preserve the setting between upgrades and will not modify any RPM installed files.

### DIFF
--- a/deploy/packaging/rpm/centos/6/SOURCES/geowave-init.sh
+++ b/deploy/packaging/rpm/centos/6/SOURCES/geowave-init.sh
@@ -49,7 +49,10 @@ wait_for_start() {
 start() {
   echo -n "Starting $DISPLAY_NAME: "
   if ! is_running; then
-  	runuser -l geowave -c "$GEOWAVE_HOME/geoserver/bin/startup.sh &" > /dev/null 2>&1
+    if [ -f /etc/geowave/geowave.config ]; then
+      SOURCE_CMD="source /etc/geowave/geowave.config;"
+    fi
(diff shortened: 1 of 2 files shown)


deploy/packaging/rpm/centos/6/SOURCES/geowave-init.sh
```diff
@@ -49,7 +49,10 @@ wait_for_start() {
 start() {
   echo -n "Starting $DISPLAY_NAME: "
   if ! is_running; then
-  	runuser -l geowave -c "$GEOWAVE_HOME/geoserver/bin/startup.sh &" > /dev/null 2>&1
+    if [ -f /etc/geowave/geowave.config ]; then
+      SOURCE_CMD="source /etc/geowave/geowave.config;"
```
chrisbennight: Does this file get manage via alternatives?
(I can see arguments for different config files per version, but then I'm not sure what the expectation is re: upgrading - copy over currently linked file?)

spohnan: @chrisbennight The config file would not be managed by RPM and if absent you'd be using the default settings. I'd probably use  Puppet to mange the file with something like the following.

```
# Bump up the memory allowed for GeoServer
  file { '/etc/geowave/geowave.config':
    ensure  => file,
    owner   => 'geowave',
    group   => 'geowave',
    mode    => 644,
    content => 'export JAVA_OPTS="-Xmx2500m -XX:MaxPermSize=128m"',
    require => Package['geowave-core'],
    notify  => Service['geowave'],
  }
```

Issue #423, opened 2015-06-09
Support Cloudera Single User Mode Installs


spohnan:
Definition of single user mode install below. They suggest naming your single user cloudera-scm in the documentation steps so I can add that to the end of the list of users our RPM script will try. The docs also specify that you *can* name your user anything you want which we can't support from RPMs. For this case of a single user mode install with a custom user name I can modify the script to accept a --user argument which can be used if you re-run the script manually after the RPM install reports a failure to upload the iterator.

[Excerpt from Cloudera Documentation](http://www.cloudera.com/content/cloudera/en/documentation/core/latest/topics/install_singleuser_reqts.html)

In a conventional Cloudera Manager deployment, the Cloudera Manager Agent, which manages Hadoop processes on each host, runs as the root user. However, some environments restrict access to the root account.

Cloudera Manager 5.3 and higher provides single user mode, which satisfies the requirements of such environments. In single user mode, the Cloudera Manager Agent and all the processes run by services managed by Cloudera Manager are started as a single configured user and group. Single user mode prioritizes isolation between Hadoop and the rest of the system over isolation between Hadoop processes running on the system.

spohnan:
CDH single user install supported as of #424

PR #424, opened 2015-06-10
Support CDH single user install

spohnan:
Addresses issue #423 and will add one additional HDFS user to the list of default usernames array. If a custom user we'll never be able to guess was used add a --user argument to the script so it doesn't have to be modified when re-run manually.


rfecher:
thanks for covering the single-user cdh install situation

Issue #425, opened 2015-06-10
Use PreparedGeometries for intersection in spatial query filters


rwgdrummer:

chrisbennight:
A bit more info - but this looks like a solid default choice
https://github.com/chrisbennight/intersection-test

PR #426, opened 2015-06-11
Issue #425 Optimize spatial queries to use prepared geometries.

rwgdrummer:
The new approach also adds an optimization for the use of Interning geometries.  The prior approach lead to excessive memory use due to the reconstitution of geometry PRIOR to interning, creating unnecessary duplicates.

### DIFF
--- a/core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
+++ b/core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
@@ -2,6 +2,7 @@ package mil.nga.giat.geowave.core.geotime.store.filter;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -21,6 +22,8 @@ import mil.nga.giat.geowave.core.store.filter.BasicQueryFilter;
 import mil.nga.giat.geowave.core.store.filter.GenericTypeResolver;
 
 import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.prep.PreparedGeometry;
+import com.vividsolutions.jts.geom.prep.PreparedGeometryFactory;
 
 /**
  * This filter can perform fine-grained acceptance testing (intersection test
@@ -31,9 +34,9 @@ public class SpatialQueryFilter extends
 		BasicQueryFilter
 {
 
-	private static final Interner<Geometry> geometryInterner = Interners.newWeakInterner();
-
-	private Geometry queryGeometry;
+	private static final Interner<GeometryImage> geometryImageInterner = Interners.newWeakInterner();
+	public static final PreparedGeometryFactory FACTORY = new PreparedGeometryFactory();
+	private GeometryImage preparedGeometryImage;
 
 	private Set<ByteArrayId> geometryFieldIds;
 
@@ -59,7 +62,8 @@ public class SpatialQueryFilter extends
 		super(
 				strippedGeometry.strippedQuery,
 				strippedGeometry.strippedDimensionDefinitions);
-		this.queryGeometry = queryGeometry;
+		preparedGeometryImage = new GeometryImage(
+				FACTORY.create(queryGeometry));
 		geometryFieldIds = strippedGeometry.geometryFieldIds;
 	}
 
@@ -115,7 +119,7 @@ public class SpatialQueryFilter extends
 	@Override
 	public boolean accept(
 			final IndexedPersistenceEncoding persistenceEncoding ) {
-		if (queryGeometry == null) {
+		if (preparedGeometryImage == null) {
 			return true;
 		}
 		// we can actually get the geometry for the data and test the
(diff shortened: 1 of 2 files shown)


core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
```diff
@@ -196,8 +204,73 @@ public void fromBinary(
...
+				geometryBinary));
+		preparedGeometryImage.init();
 
 		super.fromBinary(theRest);
 	}
+
+	/**
+	 * Used for interning Geometry. Interning a geometry provides no additional
+	 * benefit since the geometry is already constructed in the heap at the
+	 * interning begins. The byte image of geometry provides a more efficient
```
chrisbennight: "Interning a geometry provides no additional benefit since the geometry is already constructed in the heap at the interning begins"  -  I'm not quite sure the scope of the comment - do you mean interning geometries in general, interning geometries in addition to the byte[] value, etc?  As I read it I disagree with this comment - or at a minimum it doesn't accurately convey the issue.
So the point of interning was to have one instance of a geometry object maintained as iterator state per JVM (or I guess classloader) instance.   Since we could end up with tens of thousands of iterators per scan range a 2 MB geometry could explode into gigabytes of data, completely using up the head allocated to the tablet server.

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
```diff
@@ -196,8 +204,73 @@ public void fromBinary(
...
+	 * component to hash and associate with a single image of the geometry.
+	 * Prepared geometries cannot be interned since they do not extend
+	 * Object.hashCode().
+	 * 
+	 * Using this approach increased performance of a large query unit test by 40% and
+	 * eliminated out of heap memory exceptions during maven builds.
```
chrisbennight: IMO you need to clarify this comment more - I don't believe this iterator stack was benchmarked;  rather the comparison was using a preparedgeometry vs. a non prepared geometry.   Which was probably not where you were running out of heap - did that happen because you couldn't intern a prepared geometry?   I think that's two different cases though.

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
```diff
@@ -196,8 +204,73 @@ public void fromBinary(
 					fieldId));
 		}
 		buf.get(theRest);
-		queryGeometry = geometryInterner.intern(GeometryUtils.geometryFromBinary(geometryBinary));
+		// cannot intern the prepared geometry. It uses a Object.hashCode().
+		preparedGeometryImage = geometryImageInterner.intern(new GeometryImage(
+				geometryBinary));
+		preparedGeometryImage.init();
```
chrisbennight: Is there a good reason to have a separate .init() method?   Since there's no parameterless constructor for GeometryImage  you would only ever be calling init() directly after a constructor call (I don't see any benefit to deferring the init;  in that case you would just deferr object construction all together).

So thoughts on making the init() method private and calling it explicitly in the GeometryImage constructor(s)?

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
rwgdrummer: A geometry is constructed first prior to interning.  It consumes heap space before being released to its interned duplicate. This approach prevents a Geometry from being constructed if it already exists.

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
rwgdrummer: No.  The test ran out of heap when the code interned the geometry PRIOR to creating the prepared geometry.   The use of the prepared geometry required an alternate approach to interning the Geometry.  I am not sure what you mean by not being benchmarked.  This was a separate effort from the benchmark work I did yesterday.

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
rwgdrummer: The reason is to delay the construction of the Geometry.  I could have made the 'getter' synchronized, placing the code there.  However, the overhead of synchronizing at that point is unnecessary in inefficient since it is called many times during the filter evaluation.   

After obtaining the image from the interner, there are two possible conditions: the image was already interned by prior code (already initialized) or this new image was interned, requiring initialization.

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/store/filter/SpatialQueryFilter.java
```diff
@@ -196,8 +204,83 @@ public void fromBinary(
...
+
+		public GeometryImage(
+				byte[] geometryBinary ) {
+			super();
+			this.geometryBinary = geometryBinary;
+			//init();
+		}
+
+		public synchronized void init() {
+			if (preparedGeometry == null) preparedGeometry = FACTORY.create(GeometryUtils.geometryFromBinary(geometryBinary));
```
rfecher: our formatter will change this line

rwgdrummer: Will fix this prior to pull.

Issue #235, opened 2015-02-12
Migrate to use ContentDataStore


rfecher:
GeoTools has marked AbstractDataStore as deprecated and is suggesting projects use ContentDataStore instead.

rfecher:
in talking with the experts, we need to associate user authorizations with a transaction within the contentstate object, and if the transaction is auto-commit it we should use an empty set of authorizations.  ContentState caches counts and boundingbox now which is pretty cool, it also really handles transactions for us in a very reasonable way (by keeping track of a diff). For the cached count and bounding box we will have to expand to include our stats for the given user authorization (and presumably then cache that result, the cache is copied from the contentstate of transaction auto-commit but is stored on a per transaction basis so it should fit pretty well).  
For example here's the map that caches state per transaction:
https://github.com/geotools/geotools/blob/12.2/modules/library/data/src/main/java/org/geotools/data/store/ContentEntry.java#L70

rwgdrummer:
I would hold up on this until we discuss further.  I looked at the content data store in the early days of developing WFS-T.  it holds everything in memory including additions (scale issue).   The model we implemented scales better and supports the same functionality.  The transaction state does have authorizations associated with each transaction.  The only thing we do not have is the cached counts and bounding box, which is managed by the accumulo cache for statistics. When accessing statistics, the user's credentials association with the transaction:

final Map<ByteArrayId, DataStatistics<SimpleFeature>> statsMap = reader.getComponents().getDataStatistics(
					reader.getTransaction());

rwgdrummer:
Question: Is there a claim that ContentDataStore helps solve the GWC visibilities issue?

chrisbennight:
Re: the GWC issues, that's a separate thread (#217 - with some more info added there)
I don't see how the ContentDataStore would impact the GWC visibility issues (mostly the same issue as needing different tiles for different CQL parameter sets) - but I could be missing something.   

Re: authorizations + transaction - probably part of another conversation, but I"m not getting why we use an empty set of authorizations with an auto-commit transaction?

rfecher:
Re: GWC, I think you're right, different issue

Re: authorizations + transaction - basically all content states for all transactions are initialized with the content state established with an auto-commit transaction so the way I'm looking at it we want it to be the least common denominator, an empty set of authorizations

chrisbennight:
Gotcha - I had gone down a segue with default authorizations, but that turned out to be me mis-remembering some stuff.

[...]

PR #430, opened 2015-06-17
ContentDataStore

dcy2003:
Addresses #235 

GeoWaveGTDataStore now uses ContentDataStore instead of AbstractDataStore which has been deprecated.  Also, GeoWaveFeatureSource now uses ContentFeatureStore and GeoWaveGTDataStoreFactory implements DataStoreFactorySpi since both of those classes also previously inherited from deprecated classes.


dcy2003:
Hold off on merging this.  Need to remove some unnecessary logic related to heatmap that I ported over from the old data store.  Will squash down to a single commit and notify that it's ready for merging after I can verify all is well after removal.

dcy2003:
OK - this should be good to go

PR #443, opened 2015-06-30
GEOWAVE-441.  Improve performance of WFS-T with small transaction set…

rwgdrummer:
…s (e.g. < 10000) using buffer and bulk writes

### DIFF
--- a/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/transaction/GeoWaveTransactionManagement.java
+++ b/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/transaction/GeoWaveTransactionManagement.java
@@ -447,18 +447,10 @@ public class GeoWaveTransactionManagement implements
 						feature = modRecord.newFeature;
 					}
 					else if ((oldFeatures != null) && !oldFeatures.isEmpty()) {
-						// need
-						// to
-						// check
-						// if
-						// the
-						// removed
-						// feature
-						// was
-						// just
-						// moved
-						// meaning its original matches the boundaries of this
-						// 'feature'. matchesOne(oldFeatures, feature))
+						// need to check if the removed feature
+						// was just moved meaning its original matches the
+						// boundaries of this 'feature'. matchesOne(oldFeatures,
+						// feature))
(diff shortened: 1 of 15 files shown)


rwgdrummer:
This approach retains current functionality while gaining performance through batch insertions to the DataStore.  This approach does not require any additional work/configuration from the users/administrators.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/transaction/GeoWaveTransactionManagement.java
```diff
@@ -403,29 +436,31 @@ public void remove() {}
...
+						// need
+						// to
+						// check
+						// if
+						// the
+						// removed
+						// feature
+						// was
+						// just
+						// moved
```
chrisbennight: Commenting format is wonderful :) :)

chrisbennight:
Gotcha - I was confused the first time we talked I think;  I thought you were replacing the visibility based transaction isolation with just a local in memory map.    (which isn't at all what's happening)

Is it fair to say that this is jut buffering the contents of transactions in an in memory structure and only flushing them to the datastore when flush is called  - or when they have exeeded the max buffer size?

So really it would be of greatest benefit to largest transaction sets, not small (i.e. if the transaction set was ~1 it would be similar to previous behavior?)

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/transaction/GeoWaveTransactionManagement.java
rwgdrummer: Glad you like it.  So much for auto-formatting.

rwgdrummer:
The benefit to any transaction, where the transactions fit in the buffer, is efficient bulk-writes without the overhead of transforming the visibility.  This is the typical use case.  Thus, users that are writing small batches of transactions gain immediate benefit without changing their code or configuration.

The benefit to large transactions that exceed the buffer size is the efficient bulk-writes. 
All the while, transaction isolation is still maintained in those cases.

PR #445, opened 2015-06-30
Implemented a Compound Index Strategy

meislerj:
Implemented CompoundIndexStrategy. For now it allows the
composition of two NumericIndexStrategy objects into a
single NumericIndexStrategy. This can easily be modified
to support N sub-indeces. The dimensions of the
CompoundIndexStrategy are the sum of the dimensions of
the sub-indeces. The compound index prefixes the bytes
from the first sub-index before any of the bytes from the
second. Externally, the compound index can be treated like
any other multi-dimensional index.

### DIFF
--- a/core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
+++ b/core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
@@ -1,0 +1,345 @@
+package mil.nga.giat.geowave.core.index;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import mil.nga.giat.geowave.core.index.dimension.NumericDimensionDefinition;
+import mil.nga.giat.geowave.core.index.sfc.data.BasicNumericDataset;
+import mil.nga.giat.geowave.core.index.sfc.data.MultiDimensionalNumericData;
+import mil.nga.giat.geowave.core.index.sfc.data.NumericData;
+
+public class CompoundIndexStrategy implements
+		NumericIndexStrategy
+{
+
+	private String id;
+	private NumericIndexStrategy subStrategy1;
+	private NumericIndexStrategy subStrategy2;
+	private NumericDimensionDefinition[] baseDefinitions;
+	private int defaultNumberOfRanges;
+
+	public CompoundIndexStrategy(
+			final String id,
+			final NumericIndexStrategy subStrategy1,
+			final NumericIndexStrategy subStrategy2 ) {
+		this.id = id;
+		this.subStrategy1 = subStrategy1;
+		this.subStrategy2 = subStrategy2;
+		baseDefinitions = createNumericDimensionDefinitions();
+		defaultNumberOfRanges = (int) Math.ceil(Math.pow(
+				2,
+				getNumberOfDimensions()));
+	}
+
+	public NumericIndexStrategy[] getSubStrategies() {
+		return new NumericIndexStrategy[] {
+			subStrategy1,
+			subStrategy2
+		};
+	}
+
+	@Override
+	public byte[] toBinary() {
+		final byte[] idBinary = StringUtils.stringToBinary(id);
+		final byte[] delegateBinary1 = PersistenceUtils.toBinary(subStrategy1);
+		final byte[] delegateBinary2 = PersistenceUtils.toBinary(subStrategy2);
+		final ByteBuffer buf = ByteBuffer.allocate(8 + idBinary.length + delegateBinary1.length + delegateBinary2.length);
+		buf.putInt(idBinary.length);
+		buf.put(idBinary);
+		buf.putInt(delegateBinary1.length);
+		buf.put(delegateBinary1);
+		buf.put(delegateBinary2);
+		return buf.array();
+	}
+
+	@Override
+	public void fromBinary(
+			final byte[] bytes ) {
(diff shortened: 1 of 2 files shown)


chrisbennight:
Can you add a comment to the top of the class that explains what it does?  What you have in the description of the pull request would probably suffice :)

meislerj:
No problem

core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
```diff
@@ -0,0 +1,377 @@
...
+				rangeForStrategy1,
+				rangeForStrategy2);
+		return range;
+	}
+
+	@Override
+	public List<ByteArrayRange> getQueryRanges(
+			final MultiDimensionalNumericData indexedRange,
+			final int maxEstimatedRangeDecomposition ) {
+		final int maxEstRangeDecompositionPerStrategy = (int) Math.ceil(Math.sqrt(maxEstimatedRangeDecomposition));
```
rfecher: -1 is a special value that is used (anything less than 0) to mean "no max"  ...this is something we should have documented in our javadoc of the interface but I see it being used here:https://github.com/ngageoint/geowave/blob/master/core/index/src/main/java/mil/nga/giat/geowave/core/index/sfc/tiered/TieredSFCIndexStrategy.java#L43

core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
```diff
@@ -0,0 +1,377 @@
...
+		return new ByteArrayRange(
+				start,
+				end);
+	}
+
+	private List<ByteArrayRange> getByteArrayRanges(
+			final List<ByteArrayRange> ranges1,
+			final List<ByteArrayRange> ranges2 ) {
+		final List<ByteArrayRange> ranges = new ArrayList<>(
+				ranges1.size() + ranges2.size());
```
rfecher: not a big deal because the list is going to dynamically resize anyways, but if you want to set a good initial size shouldn't it be ranges1.size() * ranges2.size()?

core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
rfecher: also, I'd argue if the argument is 'maxEstimatedRangeDecomposition' that you'd take the floor instead of the ceiling to try to keep the result less than max rather than more than max, but I could see arguments either way I suppose

core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
```diff
@@ -0,0 +1,377 @@
...
+		return getInsertionIds(
+				indexedData,
+				defaultNumberOfRanges);
+	}
+
+	@Override
+	public List<ByteArrayId> getInsertionIds(
+			final MultiDimensionalNumericData indexedData,
+			final int maxEstimatedDuplicateIds ) {
+		final int maxEstDuplicatesPerStrategy = (int) Math.ceil(Math.sqrt(maxEstimatedDuplicateIds));
```
rfecher: same issue, just wondering whether we really want floor to respect "max"

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/GeoWaveGTDataStoreFactory.java
```diff
@@ -134,9 +134,8 @@ public boolean canProcess(
 			return true;
 		}
 		catch (GeoWavePluginException e) {
-			LOGGER.error(
```
rfecher: good catch - if anything this should have been at the info level, but definitely not error

rfecher:
I made a few comments but really I think the only significant one is watching out for negatives on the max query range. We didn't really well document our use of -1 as a pass-through for no max query ranges, but its used elsewhere particularly to re-use code between the 2 getQueryRange methods to not use a capped maximum query range.  So although its not directly in the call chain I think that would affect this CompoundIndexStrategy (its definitely in the tierednumericindexstrategy at least), I'd try to be consistent with its handling of negative values (and we really should have had it in the javadoc of the interface).

core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
meislerj: Agreed.

core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
meislerj: Regarding floor vs ceiling: I thought about that also.  It's a tradeoff.  If you have a compound index strategy with 3 dimensions, the default maxEstimatedRangeDecomposition is 8 (2^d).  The way this is currently done, getQueryRanges(...) for each sub-index gets called with the square root of the maxEstimatedRangeDecomposition.  Using the floor would yield ~4 ranges (depending on how closely each sub-index respects the parameter) instead of ~9 ranges.  It's a tradeoff between giving too coarse coverage of the indexedRange or returning too many ranges.  Since the javadoc said that the parameter is more of a suggestion than a strict limit, I chose ceiling.  I'm happy to switch it if you feel it makes more sense.

PR #458, opened 2015-07-16
Automate version change updates

spohnan:
A helper script to automate the find-n-replace process for version strings in pom.xml files during a version bump.


jwomeara:
Nice!

spohnan:
Will want to try and use the maven release plugin instead if possible.

Issue #464, opened 2015-07-17
CommandLine Ingest throws Exception when trying to use flags from plugin


ewilson-radblue:
First, I plan to submit a pull request with a fix for this issue, but I wanted to create an issue here just in case someone is already working on it.

The problem occurs when you try to use fields specified in an IngestFormatOptionProvider from a plugin.  In AbstractIngestCommandLineDriver, it tries to parse the command on line 108 but does not call IngestFormatOptionProvider's applyOptions() method until it calls getPluginProviders() on line 141.  It would try to reparse the command on line 162 with the additional options, but the failed parse on line 108 throws an exception and never gets to that line.

Also, if the 'f' flag is not specified, the custom options are never added at all.

PR #465, opened 2015-07-17
Fixes custom flags in commandline ingest.

ewilson-radblue:
Addresses #464 (I suppose creating that issue wasn't necessary).

The problem was when using fields specified in an IngestFormatOptionProvider from a plugin. In AbstractIngestCommandLineDriver, it tries to parse the command on line 108 but does not call IngestFormatOptionProvider's applyOptions() method until it calls getPluginProviders() on line 141. It would try to reparse the command on line 162 with the additional options, but the failed parse on line 108 throws an exception and never gets to that line.  Also, if the 'f' flag is not specified, the custom options are never added at all.

The fix is to set the parameter 'stopAtNonOption' to 'true' for CommandLineParser.parse() so that it doesn't throw an exception on an unrecognized parameter.  This allows it to pull out the 'f' flag value and apply the flags for the appropriate plugins.  Also, moved the application of the plugins' custom parameters to apply even in cases where the 'f' flag is not specified.


rfecher:
this is going to be merged with GEOWAVE-464

Issue #468, opened 2015-07-17
Extend GeoTools vector ingest to perform configurable String to Date field retyping


ewilson-radblue:
Will add a custom flag to be used to add a mapping from a field name to a Java SimpleDateFormat.  It will use commas to separate each entry, then the first ':' character will separate the field name from the format.

For example:
... -date "time:MM:dd:YYYY,time2:YYYY/MM/dd hh:mm:ss" ...
This will configure two fields for retyping, 'time' and 'time2', each expecting to convert a different format.

PR #470, opened 2015-07-17
Adds Date retyping for GeoTools Vector ingest (#468)

ewilson-radblue:
Adds Date retyping for GeoTools Vector ingest (#468).  To support it, adds CompoundIngestFormatOptionProvider to enable multiple providers for a single ingest plugin.

### DIFF
--- a/extensions/formats/geotools-vector/src/main/java/mil/nga/giat/geowave/format/geotools/vector/retyping/date/DateFieldOptionProvider.java
+++ b/extensions/formats/geotools-vector/src/main/java/mil/nga/giat/geowave/format/geotools/vector/retyping/date/DateFieldOptionProvider.java
@@ -1,0 +1,73 @@
+package mil.nga.giat.geowave.format.geotools.vector.retyping.date;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import mil.nga.giat.geowave.core.ingest.IngestFormatOptionProvider;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.log4j.Logger;
+
+public class DateFieldOptionProvider implements
+		IngestFormatOptionProvider
+{
+	private final static Logger LOGGER = Logger.getLogger(DateFieldOptionProvider.class);
+
+	private Map<String, String> fieldToFormatMap = null;
+
+	@Override
+	public void applyOptions(
+			Options allOptions ) {
+		final Option dateOption = new Option(
+				"date",
+				true,
+				"A map of date field names to the date format of the file. Use commas to separate each entry, then the first ':' character will separate the field name from the format. Use '\\,' to include a comma in the format. For example: \"time:MM:dd:YYYY,time2:YYYY/MM/dd hh:mm:ss\" configures fields 'time' and 'time2' as dates with different formats");
+		allOptions.addOption(dateOption);
+	}
+
+	@Override
+	public void parseOptions(
+			CommandLine commandLine ) {
+		if (commandLine.hasOption("date")) {
+			fieldToFormatMap = new HashMap<>();
+			final String arg = commandLine.getOptionValue("date");
+			if (arg != null) {
+				String[] values = arg.split(",");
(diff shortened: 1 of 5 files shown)


ewilson-radblue:
Only the first Travis build of the four is failing because it can't find one of the classes... but it's clearly there.  I'm not sure what the problem is.

chrisbennight:
So the failure is valid - it's coming from changes in geotools between 12.x and 13.x+;   the SimpleFeatureTypeBuilder class had a bunch of set(...) methods added here: https://github.com/geotools/geotools/commit/33e9de5a0306761950ac7dbffc428d9493389b20#diff-69cbda95a5fcc79676d0c12b73dd3f21
the diff makes use of one of those methods here:
https://github.com/ngageoint/geowave/blob/GEOWAVE-468/extensions/formats/geotools-vector/src/main/java/mil/nga/giat/geowave/format/geotools/vector/retyping/date/DateFieldRetypingSource.java#L66

-  and the first profile uses 12.4 of geotools, so the method isn't found.

Now what to do about this;  typically we have been looking to stick with N and N-1.  14.x isn't out yet.  I looking at what got added in the commit above though, and what was supported in 12.x I don't see a super clean way to do this (since attributes is protected, and there's no accessor) - but I just glanced it it real quick.   

Dropping 12.x support a bit early isn't out of the question - just want to get a double check and make sure there's not an easy way to maintain it.   @rfecher

rfecher:
that is kinda annoying - I think there should be a much less elegant way to set an attribute descriptor in 12.x that will also work in 13+ (I think looping over the attributes and just adding the attribute descriptor with SimpleFeatureTypeBuilder.add() if its not the date field and adding the new descriptor if it is the date field). But they are close to releasing v14.0 so my vote is that we drop support for 12.x in our next minor release (0.8.9).  We are releasing 0.8.8.1 on Monday with some bug fixes and Eric's other commit can make it into that but we'll hold off on this commit (so 0.8.8.1 will be the last one with geotoools 12.x support).

ewilson-radblue:
I actually had it at first where it would loop through the AttributeDescriptors of the initial type, re-add them if they weren't the ones marked to replace, and replace them if need be.  It worked but I thought this solution was more concise.

I can switch back to that if it would be preferable to also maintain GeoTools 12.x support.

extensions/formats/geotools-vector/src/main/java/mil/nga/giat/geowave/format/geotools/vector/retyping/date/DateFieldOptionProvider.java
```diff
@@ -0,0 +1,51 @@
...
+			final String arg = commandLine.getOptionValue("date");
+			if (arg != null) {
+				String[] values = arg.split(",");
+				for (String entry : values) {
+					final int firstSemiCol = entry.indexOf(':');
+					fieldToFormatMap.put(
+							entry.substring(
+									0,
+									firstSemiCol),
+							entry.substring(firstSemiCol + 1));
```
rwgdrummer: Would trim() the substring to be safe from whitespace

extensions/formats/geotools-vector/src/main/java/mil/nga/giat/geowave/format/geotools/vector/retyping/date/DateFieldRetypingSource.java
```diff
@@ -0,0 +1,99 @@
...
+
+	@Override
+	public Object retypeAttributeValue(
+			Object value,
+			Name attributeName ) {
+		Object outValue = value;
+		final String localName = attributeName.getLocalPart();
+		if (value != null && fieldNameToTimestampFormat.containsKey(localName)) {
+			try {
+				outValue = new SimpleDateFormat(
```
rwgdrummer: Knowing the SampleDateFormat is not Thread safe, this approach is an easy answer.  Since ingest is a high-performance component, I would like to see a ThreadLocal cache of pre-built SimpleDateFormats.

ewilson-radblue:
When convenient, I think it's ready to be merged.  I squashed my changes into one commit, switched back to support GeoTools 12.x, and addressed the code comments.  Thanks!

PR #474, opened 2015-07-20
Supports Min/Max Range of ROW IDs per index.

rwgdrummer:


rwgdrummer:
Wait up on this.  There is a boundary condition that is failing.

rwgdrummer:
Fixed the issues.

Issue #484, opened 2015-07-31
Remotely invoked MapReduce Jobs need to be able to specify additional JARs on classpath


ewilson-radblue:
The recommended method to do this is to to put the JARs in HDFS and add them to the Job's classpath.  Most of the work can be done by the invoker, but as the Job objects are being created within the Tool implementations, GeoWave needs an option to add the JARs to the classpath of the Job.

ewilson-radblue:
Merge fixes the problem for the KDEJobRunner, but the issue still exists for the other tools.  If they are to be remotely runnable, they should have a similar capability.

datasedai:
Isn't there a hadoop option to do this?  Like HADOOP_CLASSPATH or something?

PR #485, opened 2015-07-31
Enables KDEJobRunner to add JARs to Job classpath from configured array of HDFS paths

ewilson-radblue:
See #484

### DIFF
--- a/analytics/mapreduce/src/main/java/mil/nga/giat/geowave/analytic/mapreduce/kde/KDEJobRunner.java
+++ b/analytics/mapreduce/src/main/java/mil/nga/giat/geowave/analytic/mapreduce/kde/KDEJobRunner.java
@@ -1,6 +1,8 @@
 package mil.nga.giat.geowave.analytic.mapreduce.kde;
 
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 
 import mil.nga.giat.geowave.adapter.raster.RasterUtils;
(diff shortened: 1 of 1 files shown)


analytics/mapreduce/src/main/java/mil/nga/giat/geowave/analytic/mapreduce/kde/KDEJobRunner.java
```diff
@@ -431,4 +444,19 @@ protected int getCQLFilterArg() {
 		return 15;
 	}
 
+	protected void addJobClasspathDependencies(
+			final Job job,
+			final Configuration conf )
+			throws IOException,
+			URISyntaxException {
+		final String[] jars = conf.getTrimmedStrings(GEOWAVE_CLASSPATH_JARS);
```
rfecher: where does the value for this 'geowave.classpath.jars' get in the job configuration in the first place?

ewilson-radblue: I'm able to pass it in from the invoking application.  For example:

    final Configuration conf = new Configuration();
    final String[] hdfsPaths = new String[] {...};
    conf.setStrings(KDEJobRunner.GEOWAVE_CLASSPATH_JARS, hdfsPaths);
    
    final String[] args = new String[] {...};
    ToolRunner.run(
        conf,
        new KDEJobRunner(),
        args);

So my application will upload the JARs to HDFS, and just pass the paths in through the configuration.

PR #548, opened 2015-11-09
Some ingesting and querying use case examples.

rnix:
Just adding some examples we found useful to learn/understand.

### DIFF
--- a/examples/src/main/java/mil/nga/giat/geowave/examples/query/SpatialQueryExample.java
+++ b/examples/src/main/java/mil/nga/giat/geowave/examples/query/SpatialQueryExample.java
@@ -1,0 +1,504 @@
+package mil.nga.giat.geowave.examples.query;
+
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.Point;
+import com.vividsolutions.jts.io.ParseException;
+import com.vividsolutions.jts.io.WKTReader;
+import mil.nga.giat.geowave.adapter.vector.FeatureDataAdapter;
+import mil.nga.giat.geowave.adapter.vector.VectorDataStore;
+import mil.nga.giat.geowave.core.geotime.GeometryUtils;
+import mil.nga.giat.geowave.core.geotime.IndexType;
+import mil.nga.giat.geowave.core.geotime.store.query.SpatialQuery;
+import mil.nga.giat.geowave.core.index.ByteArrayId;
+import mil.nga.giat.geowave.core.store.CloseableIterator;
+import mil.nga.giat.geowave.datastore.accumulo.BasicAccumuloOperations;
+import mil.nga.giat.geowave.datastore.accumulo.metadata.AccumuloAdapterStore;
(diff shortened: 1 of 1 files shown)


examples/src/main/java/mil/nga/giat/geowave/examples/query/SpatialQueryExample.java
```diff
@@ -0,0 +1,497 @@
...
+		/**
+		 * 1- Adapter previously obtained from the feature name.
+		 * 2- Default spatial index.
+		 * 3- A SpatialQuery, which takes the query geometry - aka Bounding box
+		 * 4- Filters. For this example, no filter is used.
+		 * 5- Limit. Same as standard SQL limit. 0 is no limits.
+		 * 6- Accumulo authorizations. For our mock instances, "root" works. In a real Accumulo setting, whatever authorization is associated to the user in question.
+		 */
+
+		CloseableIterator<SimpleFeature> queryResultIterator = dataStore.query(
```
rwgdrummer: Place within a try(resource) {
}.
Repeat for all query examples.

examples/src/main/java/mil/nga/giat/geowave/examples/query/SpatialQueryExample.java
```diff
@@ -0,0 +1,497 @@
...
+ * 1- Querying by polygon a set of points.
+ * 2- Filtering on attributes of features using CQL queries
+ * 3- Ingesting polygons, and running polygon intersect queries.
+ * You can check all points, geometries and query accuracy in a more visual manner @ http://geojson.io/
+ */
+public class SpatialQueryExample {
+    private static Logger log = Logger.getLogger(SpatialQueryExample.class);
+
+	//We'll use GeoWave's VectorDataStore, which allows to run CQL rich queries
+	private static VectorDataStore dataStore;
```
rwgdrummer: Although you can use this interface now.  It is short lived.  It will  not be part of the API in 0.9

rfecher: to further clarify - vectordatastore provides some query methods that are specific to SimpleFeature data, such as the CQL query on line 370 of this example.  In 0.9 we try to eliminate the need for this specialized interface, the query method that you can use for CQL filtering is the same query method as in the DataStore interface, you just pass in a query object that is a CQLQuery within the vector adapter module. I'm happy with accepting this as the right example for CQL queries under 0.9-snapshot, and updating it for 0.9 as we pull in some of the broader changes (so our GEOWAVE-238-50_merge branch is actually where a lot of development is being staged for some 0.9 API changes)

examples/src/main/java/mil/nga/giat/geowave/examples/query/SpatialQueryExample.java
rfecher: I agree - for accumulodatastore for example, the result iterators end up wrapping an accumulo scanner and you really would want to close this iterator when you are finished which in turn closes the underlying scanner

mses-bly:
We will have to adapt our internal tools when the new methods for query are ready, and we'll also update the example.

PR #558, opened 2015-12-02
Addition of AvroFeatureAdapter

mrdahlb:
This new adapter is intented to serialize all the attributes of a SimpleFeature as a single object.  This is meant as an optimization when visibility is the same for all attributes of the feature.
Included is the avro schema to create the generated classes from chris bennights Import/Export project, these are necessary for avro serializing/deserializing the SimpleFeature.

### DIFF
--- a/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroAttributeRowBuilder.java
+++ b/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroAttributeRowBuilder.java
@@ -1,0 +1,39 @@
+package mil.nga.giat.geowave.adapter.vector;
+
+import mil.nga.giat.geowave.core.index.ByteArrayId;
+import mil.nga.giat.geowave.core.store.adapter.NativeFieldHandler.RowBuilder;
+import mil.nga.giat.geowave.core.store.data.PersistentValue;
+
+import org.opengis.feature.simple.SimpleFeature;
+
+/**
+ * A GeoWave RowBuilder, used internally by AbstractDataAdapter to construct
+ * rows from a set field values (in this case SimpleFeatures from a set of
+ * attribute values). This implementation simply wraps a geotools
+ * SimpleFeatureBuilder.
+ * 
+ */
+public class AvroAttributeRowBuilder implements
+		RowBuilder<SimpleFeature, Object>
+{
+	private Object object;
+
+	public AvroAttributeRowBuilder() {}
+
+	@Override
+	public SimpleFeature buildRow(
+			final ByteArrayId dataId ) {
+		SimpleFeature sf = (SimpleFeature) object;
+		return sf;
+	}
+
+	@Override
+	public void setField(
+			final PersistentValue<Object> fieldValue ) {
+		// Only interested in the relevant fields
+		if (fieldValue.getId().equals(
+				AvroFeatureAttributeHandler.FIELD_ID)) {
+			object = fieldValue.getValue();
+		}
+	}
+}
--- a/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapter.java
+++ b/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapter.java
@@ -1,0 +1,91 @@
+package mil.nga.giat.geowave.adapter.vector;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import mil.nga.giat.geowave.adapter.vector.plugin.visibility.JsonDefinitionColumnVisibilityManagement;
+import mil.nga.giat.geowave.core.index.ByteArrayId;
+import mil.nga.giat.geowave.core.store.adapter.NativeFieldHandler;
+import mil.nga.giat.geowave.core.store.adapter.NativeFieldHandler.RowBuilder;
+import mil.nga.giat.geowave.core.store.adapter.PersistentIndexFieldHandler;
+import mil.nga.giat.geowave.core.store.data.field.FieldReader;
+import mil.nga.giat.geowave.core.store.data.field.FieldVisibilityHandler;
+import mil.nga.giat.geowave.core.store.data.field.FieldWriter;
+import mil.nga.giat.geowave.core.store.data.visibility.VisibilityManagement;
+import mil.nga.giat.geowave.core.store.index.CommonIndexValue;
+
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+
+/**
+ */
+public class AvroFeatureDataAdapter extends
+		FeatureDataAdapter
+{
+
+	protected AvroFeatureDataAdapter() {}
+
+	public AvroFeatureDataAdapter(
+			final SimpleFeatureType type ) {
+		super(
+				type,
+				new ArrayList<PersistentIndexFieldHandler<SimpleFeature, ? extends CommonIndexValue, Object>>());
+	}
+
+	public AvroFeatureDataAdapter(
+			final SimpleFeatureType type,
+			final VisibilityManagement<SimpleFeature> visibilityManagement ) {
+		super(
+				type,
+				new ArrayList<PersistentIndexFieldHandler<SimpleFeature, ? extends CommonIndexValue, Object>>(),
+				null,
+				visibilityManagement);
+	}
+
+	public AvroFeatureDataAdapter(
+			final SimpleFeatureType type,
+			final List<PersistentIndexFieldHandler<SimpleFeature, ? extends CommonIndexValue, Object>> customIndexHandlers ) {
+		super(
+				type,
+				customIndexHandlers,
+				null,
+				new JsonDefinitionColumnVisibilityManagement<SimpleFeature>());
+	}
+
+	public AvroFeatureDataAdapter(
+			final SimpleFeatureType type,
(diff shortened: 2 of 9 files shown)


extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapter.java
```diff
@@ -0,0 +1,123 @@
...
+			final SimpleFeatureType type ) {
+		final List<NativeFieldHandler<SimpleFeature, Object>> nativeHandlers = new ArrayList<NativeFieldHandler<SimpleFeature, Object>>(
+				1);
+
+		nativeHandlers.add(new AvroFeatureAttributeHandler());
+		return nativeHandlers;
+	}
+
+	@Override
+	protected IndexFieldHandler<SimpleFeature, Time, Object> getTimeRangeHandler(
```
jwomeara: Do we need to overload this method?  I don't think we need the AvroAdaptorProxyFieldLevelVisibilityHandler either.  The only difference between AvroAdaptorProxyFieldLevelVisibilityHandler and AdaptorProxyFieldLevelVisibilityHandler is the type of the data adapter.  However, I don't think we need to be specific about the adapter type in the Avro visibility handler because the AvroFeatureDataAdapter is a FeatureDataAdapter.  In short, I think we can reuse the AdaptorProxyFieldLevelVisibilityHandler in this method, but if we do that, the method remains unchanged from the base implementation, so I don't think we need to overload this method at all.  Hopefully that makes sense.  If not, let me know.

core/store/src/main/java/mil/nga/giat/geowave/core/store/adapter/AvroAdapterPersistenceEncoding.java
```diff
@@ -0,0 +1,42 @@
...
+import mil.nga.giat.geowave.core.store.data.PersistentValue;
+import mil.nga.giat.geowave.core.store.index.CommonIndexValue;
+
+/**
+ * This is an implementation of persistence encoding that also contains all of
+ * the extended data values used to form the native type supported by this
+ * adapter. It does not contain any information about the entry in a particular
+ * index and is used when writing an entry, prior to its existence in an index.
+ */
+public class AvroAdapterPersistenceEncoding extends
```
jwomeara: Do we need this class?  The implementation is the same as AdapterPersistenceEncoding.  I also don't see this class being used anywhere in this pull request.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroAttributeRowBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+			final ByteArrayId dataId ) {
+		SimpleFeature sf = (SimpleFeature) idToValue.get(StringUtils.stringFromBinary(id.getBytes()));
+		return sf;
+	}
+
+	@Override
+	public void setField(
+			final PersistentValue<Object> fieldValue ) {
+		// Only interested in the relevant fields
+		if (id.equals(AvroFeatureAttributeHandler.FIELD_ID)) {
```
jwomeara: Instead of using the class variable id for the comparison here, I think we should be comparing fieldValue.getId() to AvroFeatureAttributeHandler.FIELD_ID.  The class variable id is set when the AvroAttributeRowBuilder is constructed, and it looks like AvroFeatureAttributeHandler.FIELD_ID is passed, so this check will always evaluate to true.  I would just eliminate the id member variable as I don't think it's really needed.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroAttributeRowBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+ * A GeoWave RowBuilder, used internally by AbstractDataAdapter to construct
+ * rows from a set field values (in this case SimpleFeatures from a set of
+ * attribute values). This implementation simply wraps a geotools
+ * SimpleFeatureBuilder.
+ * 
+ */
+public class AvroAttributeRowBuilder implements
+		RowBuilder<SimpleFeature, Object>
+{
+	private final HashMap<String, Object> idToValue = new HashMap<String, Object>();
```
jwomeara: Does this need to be a hashmap?  We're only expecting a single fieldValue to be set (the one which represents the entire serialized Simple Feature).

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapter.java
```diff
@@ -0,0 +1,123 @@
...
+	@Override
+	public FieldWriter<SimpleFeature, Object> getWriter(
+			final ByteArrayId fieldId ) {
+		return (FieldWriter<SimpleFeature, Object>) new AvroFeatureWriter();
+	}
+
+	@Override
+	protected RowBuilder<SimpleFeature, Object> newBuilder() {
+		return new AvroAttributeRowBuilder(
+				AvroFeatureAttributeHandler.FIELD_ID);
```
jwomeara: If you make the suggested changes to the AvroAttributeRowBuilder, you won't need to pass the field id here.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/visibility/AvroAdaptorProxyFieldLevelVisibilityHandler.java
```diff
@@ -0,0 +1,64 @@
...
+ * a {@link FeatureDataAdapter}. The setup of the handlers in
+ * {@link FeatureDataAdapter} occurs prior to the resolution of the
+ * {@link VisibilityManagement}. Referencing the {@link FeatureDataAdapter}
+ * removes the initialization order constraint.
+ * 
+ * 
+ * 
+ * @param <CommonIndexValue>
+ */
+public class AvroAdaptorProxyFieldLevelVisibilityHandler implements
```
jwomeara: I commented on this above, but generally I don't think it's needed.

extensions/adapters/vector/src/test/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapterTest.java
```diff
@@ -0,0 +1,583 @@
...
+				new SpatialQuery(
+						new GeometryFactory().createPolygon(coordArray)));
+
+		int numReturned = 0;
+		while (itr.hasNext()) {
+			numReturned++;
+
+			final SimpleFeature feat = itr.next();
+
+			System.out.println("id: [" + feat.getID() + "]");
```
jwomeara: This is a good debug test for a developer, but If we're going to use this as a unit test, we should probably remove prints to standard out and instead put in some assertions to validate the results of our query.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapter.java
mrdahlb: makes sense, I created this before extending FeatureDataAdapter.  I will remove the override and the unnecessary avro handler class.

core/store/src/main/java/mil/nga/giat/geowave/core/store/adapter/AvroAdapterPersistenceEncoding.java
mrdahlb: removed

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroAttributeRowBuilder.java
mrdahlb: replaced id with fieldValue.getId()

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroAttributeRowBuilder.java
mrdahlb: removed hashma, replaced with the actual PersistentValue.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapter.java
mrdahlb: removed

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/visibility/AvroAdaptorProxyFieldLevelVisibilityHandler.java
mrdahlb: removed unnecessary class

extensions/adapters/vector/src/test/java/mil/nga/giat/geowave/adapter/vector/AvroFeatureDataAdapterTest.java
mrdahlb: changed print statements to asserts

jwomeara:
Looks good to me.

jwomeara:
Looks good to me.

PR #575, opened 2016-01-18
GEOWAVE-574: ingest into multiple dimensionality types with one command

rfecher:


core/ingest/src/main/java/mil/nga/giat/geowave/core/ingest/IngestUtils.java
```diff
@@ -62,23 +84,213 @@
...
-					indexWriter = ingestRunData.getIndexWriter(index);
+					indexWriter.write(
+							adapter,
+							geowaveData.getValue());
+				}
+			}
+		}
+	}
+
+	public static PrimaryIndex getIndex(
```
rwgdrummer: Was thinking that we could have an option to select an existing index.
Is it easier to include a custom provider for specific index or is it easier to pre-create a custom index via some index creation tool with many options per index type.  Perhaps we create a ticket for this later option since it seems like it is a bit much for the scope of this ticket.

rfecher: I kinda want a tool that can be used to recommend an index configuration.  The minimum requirement for this would be a representative set of queries, in which case we could use SFC cardinality ratios much like the best fit query planner you worked to recommend an SFC config that constrains the query the most (using a cardinality estimate similar to the choosebestindex estimator).  Additionally, as a more precise option you could give a set of files that would hypothetically be ingested, but rather than ingest them, we could just run them through some form of histogram statistic that can give good estimates for counts that match the key range represented by a query constraint for indices to inform the choice of index...

really the first route seems pretty straightforward, the second with the included histogram is a bit more complex

PR #580, opened 2016-01-19
GEOWAVE-576: added spatial subsampling to general query options

rfecher:


core/index/src/main/java/mil/nga/giat/geowave/core/index/IndexUtils.java
```diff
@@ -21,4 +21,42 @@ public static MultiDimensionalNumericData getFullBounds(
...
+		double result = Long.MAX_VALUE;
+		if (dataRangePerDimension.length == 0) {
+			return 0;
+		}
+		final double cellRangePerDimension[] = indexStrategy.getHighestPrecisionIdRangePerDimension();
+		final double bitsPerDimension[] = getBitsPerDimension(
+				indexStrategy,
+				cellRangePerDimension);
+		for (int d = 0; d < dataRangePerDimension.length; d++) {
+			final double bitsUsed = bitsPerDimension[d] - log2(Math.ceil(dataRangePerDimension[d] / cellRangePerDimension[d]));
```
rwgdrummer: The prior approach created a ratio bits required over bits available.   Minimizing the ratio was preference to use the least number of bits.  This, I believe, was aligned to the idea we discussed of looking at the number bits from the left..
This approach states that the result is bits used.  However, it seems to me this is bit unused: the total bits subtracting the bits required.  I would think that this should be maximized.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/index/ChooseHeuristicMatchIndexQueryStrategy.java
```diff
@@ -56,12 +54,15 @@ public boolean hasNext() {
...
+									dataRangePerDimension[d] = qr.getMaxValuesPerDimension()[d] - qr.getMinValuesPerDimension()[d];
+								}
+								totalMax += IndexUtils.getDimensionalBitsUsed(
 										nextIdx.getIndexStrategy(),
-										nextIdx.getIndexModel(),
-										qr);
+										dataRangePerDimension);
 							}
-							if (totalMax < min) {
+							if (totalMax > min) {
```
rwgdrummer: See prior comment.  This max makes sense to me IF we are trying to maximize the unused bits.
These reads: i want to maximize bits used.  But the most constraining dimension is going to least amount of bits.  Perhaps I am missing something.

PR #589, opened 2016-01-25
Geowave 422: update to Accumulo 1.7

rfecher:


rfecher:
I tried to credit @chrisbennight with his original commits...infamous for not squashing commits ;)

that's why this one is not squashed to one commit, I figure we can let it slide?

PR #590, opened 2016-01-25
GEOWAVE-415: update to geotools 14.1

rfecher:
geotools 14.2 is out but geoserver 2.8.2 is not, so I stuck with this release for now, the updates to new minor releases typically don't involve api changes anyways.


extensions/adapters/raster/src/main/java/mil/nga/giat/geowave/adapter/raster/RasterUtils.java
```diff
@@ -934,30 +934,37 @@ else if (sourceIsFloat) {
...
+			// categories[0] = new Category(
+			// name[b],
+			// null,
+			// targetRange,
+			// LinearTransform1D.IDENTITY);
+			// }
+			// dst[b] = new GridSampleDimension(
+			// name[b],
+			// categories,
+			// null);
```
dcy2003: Should probably either rip this commented code out entirely, OR put a FIXME in here with some notes about why it's commented out so we remember why/when to put it back eventually...

rfecher: good point, let's rip it out...tests are passing and we have revision history to go back to if for some reason we run into a snag later on

PR #596, opened 2016-01-28
GEOWAVE-593: compute stats for a query

rfecher:


extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/StatisticComputingIterator.java
```diff
@@ -0,0 +1,268 @@
...
+
+	@Override
+	public void init(
+			final SortedKeyValueIterator<Key, Value> source,
+			final Map<String, String> options,
+			final IteratorEnvironment env )
+			throws IOException {
+		try {
+			final String statsStr = options.get(STATISTICS_OPTION_NAME);
+			statsBytes = ByteArrayUtils.byteArrayFromString(statsStr);
```
rwgdrummer: statsBytes not used?

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/StatisticComputingIterator.java
```diff
@@ -0,0 +1,268 @@
...
+										// strategy and only the common index
+										// model to decode which is the case for
+										// feature data,
+										// we pass along a null strategy to
+										// eliminate the necessity to send a
+										// serialization of the strategy in the
+										// options of this iterator
+								model));
+				if (row != null) {
+					final List<ByteArrayId> insertionIds = new ArrayList<ByteArrayId>();
```
rwgdrummer: ?so we are just counting # of duplicates...unique insertion ids

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/AccumuloConstraintsQuery.java
```diff
@@ -33,12 +43,15 @@
 	protected final List<DistributableQueryFilter> distributableFilters;
 	protected boolean queryFiltersEnabled;
 
+	protected final Pair<DataAdapter<?>, DataStatistics<?>> computeStats;
```
rwgdrummer: Could we instead make a decorator class instead of injecting to this class?

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/StatisticComputingIterator.java
rfecher: this is just needed to satisfy the datastatistics interface.  Duplicates will be interesting for this interface, but ultimately I'd like this to be a different interface - call it aggregations or something
but you do want to be able to reuse existing datastatistics so they'd all have to implement an aggregation interface or whatever this is called

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/AccumuloConstraintsQuery.java
rfecher: maybe, but we do want to be able to support this for all *query* AccumuloConstraintsQueries, not delete though

Issue #583, opened 2016-01-21
Upgrade JTS version to 1.14 once it is released


dcy2003:
JTS 1.14 fixes some important thread-safety issues.

Once we perform this upgrade, we should also be able to back out of the changes put in place for #564 which compensates for a multi-threading issue related to JTS intersects() for line strings.

dcy2003:
Fixed with #605

Issue #564, opened 2016-01-09
Intermittent failure in GeoWaveBasicIT (0.9.0-RC1)


dcy2003:
Travis Build: https://travis-ci.org/ngageoint/geowave/jobs/101144205
Raw Output: https://s3.amazonaws.com/archive.travis-ci.org/jobs/101144205/log.txt

```
Tests run: 20, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 951.121 sec <<< FAILURE! - in mil.nga.giat.geowave.test.GeoWaveITSuite
testIngestAndQuerySpatialPointsAndLines(mil.nga.giat.geowave.test.GeoWaveBasicIT)  Time elapsed: 74.456 sec  <<< FAILURE!
java.lang.AssertionError: expected:<11110> but was:<11109>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at mil.nga.giat.geowave.test.GeoWaveBasicIT.testQuery(GeoWaveBasicIT.java:822)
	at mil.nga.giat.geowave.test.GeoWaveBasicIT.testIngestAndQuerySpatialPointsAndLines(GeoWaveBasicIT.java:143)


Results :

Failed tests: 
  GeoWaveBasicIT.testIngestAndQuerySpatialPointsAndLines:143->testQuery:822 expected:<11110> but was:<11109>

Tests run: 20, Failures: 1, Errors: 0, Skipped: 0
```

dcy2003:
Additional information:

Subsequent executions of this integration test reveal erratic behavior.  I have observed a mix of:
* Test Passes
* Test fails because spatial-query-with-box-filter expects 178 but returns 177 results
* Test fails because spatial-query-with-polygon-filter expects 11110 but returns 11109 results
* Test fails because spatial-temporal-query-with-polygon-filter expects 130 but returns 129 results

So far, in all failure cases, the first expected result is missing.

For example, for the spatial-query-with-box-filter, the failure case differs from the success case in that only the first result from the success case is missing from the failure case.  All other results match one to one and in order.  Here is the missing result for the spatial-query-with-box-filter failure case:

featureId:
```
tornado_tracks.7583
```

geometry:
```
[...]

PR #605, opened 2016-02-02
Upgrade to JTS v1.14

dcy2003:
Fix for #583.  Also backs out of temporary fix for #564.  I used the [FilterTest](https://github.com/ngageoint/geowave/files/98162/FilterTest.java.txt) provided by @datasedai to verify.

### DIFF
--- a/core/geotime/pom.xml
+++ b/core/geotime/pom.xml
@@ -11,7 +11,7 @@
 	<dependencies>
 		<dependency>
 			<groupId>com.vividsolutions</groupId>
-			<artifactId>jts</artifactId>
+			<artifactId>jts-core</artifactId>
 		</dependency>
 		<!--we can switch the dependency from store to ingest when accumulo becomes 
 			a pluggable data store and is no longer a dependency on ingest -->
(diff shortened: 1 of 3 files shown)


rfecher:
my one concern is regarding conflicting with geotools' dependencies:
because the artifact name changes, I imagine we'd be bringing in both dependencies? Did 1.14 change package naming conventions to avoid a conflict or are we ending up relying on the order of class loading?  If so, and it so happens that 1.14 classes are used instead of 1.13 could the 1.14 classes potentially cause runtime issues with geotools?

dcy2003:
Good catch.  Let this PR sit and I will look into this.

dcy2003:
Okay @rfecher I will let you decide which way we go here, but I pushed another commit that explicitly excludes JTS 1.13 as a transitive dependency.  I used the Maven dependency plugin to ensure that the only JTS dependencies in our entire tree are JTS 1.14.  This will ensure that anyone who relies on JTS loads the 1.14 class (GeoTools/Geoserver).  JTS 1.14 does not introduce any breaking API changes, only bug fixes and functionality/performance enhancements so this approach should be safe.  Let me know what you think.

PR #622, opened 2016-02-11
GEOWAVE-620.  Fix GeoServer pluggin WFS-T startIndex query constraint

rwgdrummer:


extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/GeoWaveFeatureCollection.java
```diff
@@ -247,7 +247,15 @@ private Filter getFilter(
...
 			timeBounds = getBoundedTime(query);
-			final Integer limit = getLimit(query);
+			Integer limit = getLimit(query);
+			final Integer startIndex = getStartIndex(query);
+			
+			// limit becomes a 'soft' constraint since GeoServer will inforce the limit
+			final Long max =  (limit != null) ? 
+					  limit.longValue() + (startIndex == null ? 0 : startIndex.longValue()) : null;
+			// limit only used if less than an integer max value.
+			limit  = (max != null && max.longValue() > Integer.MAX_VALUE) ? max.intValue() : null;
```
dcy2003: shouldn't this be ```limit  = (max != null && max.longValue() < Integer.MAX_VALUE) ? max.intValue() : null;``` (less than Integer.MAX_VALUE, not greater than) ?

dcy2003:
Suggest adding a test for this to one of the existing Integration Tests.  If nothing else, should prevent us from unknowingly breaking it in the future.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/GeoWaveFeatureCollection.java
rfecher: I'm sorry I merged and didn't catch this...

I'm pretty sure you're right.

PR #630, opened 2016-02-18
GEOWAVE-618: Adjust the index choosing heuristic to incorporate binning

rwgdrummer:


test/src/main/java/mil/nga/giat/geowave/test/GeoWaveTestEnvironment.java
```diff
@@ -230,6 +230,8 @@ public static void setup()
 						Assert.fail("Unable to start mini accumulo instance: '" + e.getLocalizedMessage() + "'");
 					}
 				}
+				// System.out.println(accumuloInstance);
```
rfecher: we should clean this up (at least at some point)

rwgdrummer: If you want, I think it would be great to just put it into a logger trace.

PR #631, opened 2016-02-18
GEOWAVE-626: Address performance issue when querying multiple range c…

rwgdrummer:
…onstraints.

### DIFF
--- a/core/index/src/main/java/mil/nga/giat/geowave/core/index/ByteArrayRange.java
+++ b/core/index/src/main/java/mil/nga/giat/geowave/core/index/ByteArrayRange.java
@@ -1,5 +1,6 @@
 package mil.nga.giat.geowave.core.index;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
@@ -92,7 +93,14 @@ public class ByteArrayRange implements
 		return ((getStart().compareTo(other.getEnd())) <= 0 && (getEnd().compareTo(other.getStart())) >= 0);
 	}
 
-	public ByteArrayRange merge(
+	public ByteArrayRange intersection(
+			ByteArrayRange other ) {
+		return new ByteArrayRange(
+				this.start.compareTo(other.start) <= 0 ? other.start : this.start,
+				this.end.compareTo(other.end) >= 0 ? other.end : this.end);
+	}
+
+	public ByteArrayRange union(
 			ByteArrayRange other ) {
 		return new ByteArrayRange(
 				this.start.compareTo(other.start) <= 0 ? this.start : other.start,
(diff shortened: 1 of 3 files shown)


core/index/src/main/java/mil/nga/giat/geowave/core/index/ByteArrayRange.java
```diff
@@ -108,26 +109,28 @@ public int compareTo(
...
-					j--;
-					ranges.set(
-							i,
-							r1);
+				}
+				else {
+					break;
 				}
 			}
+			i = j;
```
rfecher: logically this seems to have a potential inconsistency, right?  let's say you have ranges, a,b,c, and d (starts of the range are sorted in that order).  Imagine a, b, and c are very long spans, d is a very short span.  a is so long it overlaps d, but b and c aren't quite as long but overlap each other.  It seems this logic would find that a and d overlap, and skip merging b and c together.  I know this may be an edge case that we're willing to trade performance for algorithmic accuracy here.  I just wanted to at least point it out, so if we decide the accuracy tradeoff is still worth it, we at least add a comment to explain.

core/index/src/main/java/mil/nga/giat/geowave/core/index/ByteArrayRange.java
```diff
@@ -108,26 +109,28 @@ public int compareTo(
 				other.getEnd());
 	}
 
-	public static final void mergeIntersections(
+	public static final List<ByteArrayRange> mergeIntersections(
 			List<ByteArrayRange> ranges,
 			int maxRanges ) {
```
dcy2003: Seems maxRanges is unused?

core/index/src/main/java/mil/nga/giat/geowave/core/index/ByteArrayRange.java
rwgdrummer: Are you saying a does not overlap b and c?
if a does not overlap b and c, and 'a' overlaps d, then d must start before b and c.
Since the order is a,b,c,d and a overlaps d, a must have an 'end' past d's start time. The d starts after b and c, so a overlaps b and c as well.

core/index/src/main/java/mil/nga/giat/geowave/core/index/ByteArrayRange.java
rwgdrummer: Yes.  I am torn on this. 
The algorithm needs to score range distance (between each other) and size, choosing the closest/shortest ranges first until the maxRange is met.  A priority heap ordered by a size/distance metric or a sorted list ordered (in reverse) for optimal removal is needed.  Both are O(nlogn).  If you have a O(n) solution, then lets fix this now.  Otherwise, I am tempted to just drop it.  Given some maxRange (e.g. 100), a worst case scenario is probably something like three or four times this amount.

core/index/src/main/java/mil/nga/giat/geowave/core/index/ByteArrayRange.java
rfecher: got it, sorry for the confusion...`a` will merge with `b`, then `c`, then `d` in my scenario so `b` and `c` will never need to be merged directly together, ignore my comment

Issue #615, opened 2016-02-04
Server-Side (Iterator) Attribute Subsetting for Queries


dcy2003:
The recent optimization to flatten attributes with shared visibility into a single CQ broke the original implementation of supporting subsets of attributes server-side (within an iterator) for queries.  This functionality should be re-implemented.  

This ticket depends on #563 

See original ticket #143

rfecher:
take a look at [org.geotools.data.Query](https://github.com/geotools/geotools/blob/master/modules/library/api/src/main/java/org/geotools/data/Query.java#L181) to make sure that when a subset of property names are used within our geotools plugin, that the subset is passed along to geowave QueryOptions to properly subset the attribute

PR #663, opened 2016-03-11
Geowave 615

dcy2003:
Fix for #615

SharedVisibilitySplittingIterator decomposes composite fieldIds on the tablet server, which replaces the old approach of doing this twice (once client side in AccumuloUtils and also in QueryFilterIterator)

FieldFilter filters only those fieldIds that were specified in QueryOptions

AttributesSubsetQueryIT tests both server side and client side filtering functionality

We also support subsetting through GeoTools Query via propertyNames.  GeoToolsAttributesSubsetTest verifies this functionality

### DIFF
--- a/core/store/src/main/java/mil/nga/giat/geowave/core/store/query/QueryOptions.java
+++ b/core/store/src/main/java/mil/nga/giat/geowave/core/store/query/QueryOptions.java
@@ -55,7 +55,7 @@ public class QueryOptions implements
 	/**
 	 *
 	 */
-	private static final long serialVersionUID = 544085046847603372L;
+	private static final long serialVersionUID = 544085046847603371L;
 
 	private static ScanCallback<Object> DEFAULT_CALLBACK = new ScanCallback<Object>() {
 		@Override
@@ -82,6 +82,7 @@ public class QueryOptions implements
 	private transient ScanCallback<?> scanCallback = DEFAULT_CALLBACK;
 	private String[] authorizations = new String[0];
 	private Boolean dedupAcrossIndices = null;
+	private List<String> fieldIds = Collections.emptyList();
 
 	public QueryOptions(
 			final ByteArrayId adapterId,
@@ -159,6 +160,17 @@ public class QueryOptions implements
 		setLimit(limit);
 		this.scanCallback = scanCallback;
 		this.authorizations = authorizations;
+	}
+
+	/**
+	 * 
+	 * @param fieldIds
+	 *            the subset of fieldIds to be included with each query result
+	 */
+	public QueryOptions(
+			final List<String> fieldIds ) {
+		super();
+		this.fieldIds = fieldIds;
 	}
 
 	public QueryOptions() {}
@@ -357,8 +369,24 @@ public class QueryOptions implements
 		return ids;
 	}
 
-	public static long getSerialversionuid() {
-		return serialVersionUID;
+	/**
+	 * 
+	 * @return the fieldIds
+	 */
+	public List<String> getFieldIds() {
+		return fieldIds;
+	}
+
+	/**
+	 * 
+	 * @param fieldIds
+	 *            the desired subset of fieldIds to be included in query results
+	 */
+	public void setFieldIds(
+			final List<String> fieldIds ) {
+		if (fieldIds != null) {
+			this.fieldIds = fieldIds;
+		}
 	}
 
 	@Override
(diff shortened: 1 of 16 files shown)


core/store/src/main/java/mil/nga/giat/geowave/core/store/query/QueryOptions.java
```diff
@@ -379,7 +407,27 @@ public static long getSerialversionuid() {
...
+					sBuf.append(comma);
+				}
+				sBuf.append(fieldId);
+			}
+			fieldIdsBytes = StringUtils.stringToBinary(sBuf.toString());
+			fieldIdsBytesNeeded = 4 + fieldIdsBytes.length;
+		}
+
+		final ByteBuffer buf = ByteBuffer.allocate(20 + authBytes.length + aSize + iSize + fieldIdsBytesNeeded);
+		buf.putInt((fieldIdsBytesNeeded > 0) ? 1 : 0);
```
rwgdrummer: Do we need the extra int, or can we just place a 0 if the field id list is emty?

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/GeoWaveFeatureReader.java
```diff
@@ -260,13 +263,17 @@ public BaseIssuer(
 		public CloseableIterator<SimpleFeature> query(
 				final PrimaryIndex index,
 				final mil.nga.giat.geowave.core.store.query.Query query ) {
+			final QueryOptions queryOptions = new QueryOptions(
```
rwgdrummer: Could simplifying this by having a new constructor that takes a List<String> for the subset attributes where getSubset() can return an empty list when subseting is not requested?

core/store/src/main/java/mil/nga/giat/geowave/core/store/query/QueryOptions.java
```diff
@@ -379,7 +407,27 @@ public static long getSerialversionuid() {
 			}
 		}
 
-		final ByteBuffer buf = ByteBuffer.allocate(16 + authBytes.length + aSize + iSize);
+		int fieldIdsBytesNeeded = 0;
```
rwgdrummer: You used org.apache.commons.lang3.StringUtils.join else where.  Could we use that here?

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/FieldFilter.java
```diff
@@ -0,0 +1,103 @@
...
+			if (Arrays.equals(
+					k.getColumnQualifierData().getBackingArray(),
+					fieldId.getBytes())) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public static IteratorSetting getIteratorSetting() {
```
rwgdrummer: Clean...we should use the approach else where.

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/FieldFilter.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.accumulo.core.iterators.IteratorEnvironment;
+import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
+
+import com.google.common.base.Splitter;
+
+public class FieldFilter extends
+		Filter
+{
+	// must execute prior to WholeRowIterator/QueryFilterIterator but after
+	// SharedVisibilitySplittingIterator
```
rwgdrummer: Do we have an issue where subset fields may not include fields used in the filter?

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/SharedVisibilitySplittingIterator.java
```diff
@@ -0,0 +1,102 @@
...
+			throws IOException {
+
+		while (input.hasTop()) {
+
+			final Key currKey = input.getTopKey();
+			final Value currVal = input.getTopValue();
+
+			// check to see if the current value is a composite field containing
+			// multiple attributes with shared visibility
+			if (Arrays.equals(
```
rfecher: we should go forward with this PR, but we will need to benchmark it at some point - particularly relating to the use of this transforming iterator (I'm not saying it'll turn out poor performance, just making sure we at some point quantify it).

rwgdrummer:
Agreed.  Once we address these comments  We can move forward.

Steve and I were discussing potentially make a variant to the whole feature data adapter further reduce the volume from the server to the client.  

At some point, all of us did discuss a better method to compress the field/value bytes, using a bitmask--bits coincide with feature attribute names as they are in ordered within the simplefeaturetype). 

I think the next is benchmark this approach and then make the incremental changes to determine they value they provide.

core/store/src/main/java/mil/nga/giat/geowave/core/store/query/QueryOptions.java
dcy2003: Don't need the extra int.  I will clean this up.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/GeoWaveFeatureReader.java
dcy2003: Certainly could, though I'm personally of the opinion it's better to have constructors covering the required fields and setters for optional fields vs. a ton of permutations of constructors.

core/store/src/main/java/mil/nga/giat/geowave/core/store/query/QueryOptions.java
dcy2003: Yep, will do.  Will be more concise.

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/SharedVisibilitySplittingIterator.java
dcy2003: Agreed.  Said the exact same thing to Steve on Friday.  Will be curious to see what impact it has.

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/FieldFilter.java
dcy2003: I don't think we have an issue, but let's discuss to make sure I'm not missing something.

PR #674, opened 2016-03-18
GEOWAVE-673: added index as optional parameter to KDE

rfecher:


rwgdrummer:
So...move this into 0.9.0.2
Do we have a contribution to namespaces as well?

rfecher:
I didn't think we needed an 0.9.0.2 release but the formats setting namespace explicitly should be added to the 0.9.1 branch

PR #675, opened 2016-03-19
Issue 668. Removed feature namespace URI from the plugin

rwgdrummer:


rwgdrummer:
Merge and formatting makes  this look worse than it is.   Recommend merging.

Issue #660, opened 2016-03-08
Store required data statistics for discovering what adapters are in each index


rfecher:
`Map<Index,Adapter[]>` basically needs to be around to make decisions on which index to query if multiple indices exist.  This required metadata should be stored and used in place of of RowRangeHistogram which is causing this issue currently: #655

rfecher:
actually the method that would likely be used more often than not is getIndices(ByteArrayId adapterId) or something like a `Map<AdapterId, IndexIds[]>` structure

rwgdrummer:
Should apply this in the delete case as 'check' do not want to have the case where a feature is deleted in one index and not another.

Issue #659, opened 2016-03-08
Eliminate VisibilityWriter from DataStore.createIndexWriter


rfecher:
Relating to #658 if we're changing the DataStore api, we should clean this up.  The IndexWriter can be passed a VisibilityWriter on write() and that should be used instead.

Issue #658, opened 2016-03-08
DataStore createWriter to write to multiple indices


rfecher:
currently the createIndexWriter is called per index from the ingest framework and the outputformat.  It'd be cleaner to call createWriter and give it multiple indices for the datastore api.  This will also help with the callbacks to statisitics such as #652 and #655

PR #676, opened 2016-03-19
Issue #659, #658, #660.  Allow the index writer to to multiple indice…

rwgdrummer:
g


core/mapreduce/src/main/java/mil/nga/giat/geowave/mapreduce/input/GeoWaveInputFormat.java
```diff
@@ -136,7 +136,7 @@ public static DataStatisticsStore getJobContextDataStatisticsStore(
 				context);
 	}
 
-	public static void addIndex(
+	public static void setIndex(
```
rfecher: Why change it to `setIndex`?  It looks like it still calls `addIndex` under the hood and it looks like you can still have multiple indices?  Am I missing something?

core/mapreduce/src/main/java/mil/nga/giat/geowave/mapreduce/output/GeoWaveOutputFormat.java
```diff
@@ -281,7 +284,7 @@ public OutputCommitter getOutputCommitter(
 	protected static class GeoWaveRecordWriter extends
 			RecordWriter<GeoWaveOutputKey, Object>
 	{
-		private final Map<ByteArrayId, IndexWriter> indexWriterCache = new HashMap<ByteArrayId, IndexWriter>();
+		private final Map<ByteArrayId, IndexWriter> adapterWriterCache = new HashMap<ByteArrayId, IndexWriter>();
```
rfecher: we are still calling them index writers - I could understand changing it to adapter writer, but that could be up for debate.  This one isn't a big deal, just a variable name, but its seems inconsistent, and worth questioning "do we want to call these writers adapter writers?"

core/mapreduce/src/main/java/mil/nga/giat/geowave/mapreduce/input/GeoWaveInputFormat.java
rfecher: Well, I looked a little deeper, it looks while we append an index every time you call setIndex, we just seemingly arbitrarily use the first index to form query options.  Some of this was from before this pull request, but maybe something we should clean up?

core/store/src/main/java/mil/nga/giat/geowave/core/store/adapter/statistics/StatisticsProvider.java
```diff
@@ -22,4 +22,6 @@
 
 	public EntryVisibilityHandler<T> getVisibilityHandler(
 			ByteArrayId statisticsId );
+
+	public DataAdapter<T> getDataAdapter();
```
rfecher: getDataAdapter is not called anywhere from what I can tell, not sure we need it on the interface?

core/store/src/main/java/mil/nga/giat/geowave/core/store/DataStoreCallbackManager.java
```diff
@@ -10,48 +11,49 @@
...
+import mil.nga.giat.geowave.core.store.adapter.statistics.StatisticsProvider;
 import mil.nga.giat.geowave.core.store.adapter.statistics.StatsCompositionTool;
 import mil.nga.giat.geowave.core.store.index.PrimaryIndex;
 import mil.nga.giat.geowave.core.store.index.SecondaryIndexDataAdapter;
 import mil.nga.giat.geowave.core.store.index.SecondaryIndexDataManager;
 import mil.nga.giat.geowave.core.store.index.SecondaryIndexDataStore;
 
-public class DataStoreCallbackManager implements
-		AutoCloseable
+public class DataStoreCallbackManager
```
rfecher: why the change from autocloseable to now using a flush?

PR #685, opened 2016-03-29
GEOWAVE-684: export vector data commandline tools; also seems to have…

rfecher:
… formatted a few files that were in a conflicting format


rfecher:
this relates to PR #695

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/export/VectorMRExportCommand.java
```diff
@@ -0,0 +1,208 @@
...
+			final Index index = mrOptions.getIndexStore().getIndex(
+					new ByteArrayId(
+							mrOptions.getIndexId()));
+			if (index == null) {
+				JCommander.getConsole().println(
+						"Unable to find index '" + mrOptions.getIndexId() + "' in store");
+				return -1;
+			}
+			if (index instanceof PrimaryIndex) {
+				options.setIndex((PrimaryIndex) index);
```
rwgdrummer: Looking at the GeoWaveInputFormat, I think we have an issue.
I like this approach.  But, it looks like GeoWaveInputFormat overrides this selection.
I think we need a separate ticket to make sure that QueryOptions is not overridden by GeoWaveInputFormat.setIndex and setAdapters.   Basically, the proposal is to have getAdapters  and getIndex on GeoWaveInputFormat used ONLY if the QueryOptions adapters and index  are not set.  Agreed?

rwgdrummer:
Once you see the comment...I am happy with this PR.  All good.

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/export/VectorMRExportCommand.java
rfecher: good catch - that is an odd one...your proposed approach would be an improvement, or we could just rely on QueryOptions only...do you see a reason to setIndex or setAdapters on GeoWaveInputFormat separately? My assumption is that the methods in GeoWaveInputFormat pre-date QueryOptions and that is the reason they exist separately right now. Perhaps we *could* leave the setter methods around and just have it set the queryoptions?  Or I would consider removing those methods altogether and using queryoptions only unless there's something I'm missing?

rwgdrummer: Lets remove them.  I will merge this request in today.  I create a separate ticket for this additional adjustment.

PR #695, opened 2016-03-30
GEOWAVE-686: added an ingest format for avro encoded using the geowav…

rfecher:
…e simplefeaturecollection schema


rfecher:
this is branched off GEOWAVE-684 so either this can be pulled into GEOWAVE-684 first and then GEOWAVE-684 can be pulled into 0.9.1, or I can resubmit this PR into 0.9.1 if GEOWAVE-684 is first pulled into 0.9.1

PR #698, opened 2016-03-31
fixed an issue with winutils.exe being copied twice for ITs on windows

rfecher:


rfecher:
this is really basic, the code block that was deleted already exists in MiniAccumuloClusterFactory.newAccumuloCluster and is redundant so this is simply cleaning up

Issue #635, opened 2016-02-20
cli completion integration with plugins


spohnan:
The existing cli completion is a static file put into place when installing via RPM and works as long as all the possible arguments are known in advance.

https://github.com/ngageoint/geowave/blob/master/core/cli/src/main/resources/geowave-tools-cmd-completion.sh

An example of how this could be made aware of non-default plugin options or even more dynamic environmental options can be seen in the command completion module for git.

https://github.com/git/git/blob/v2.7.1/contrib/completion/git-completion.bash#L404

I'm not advocating we need to maintain a 2500+ line bash script just for command line completion but some of the concepts illustrated in the git script could be used to make the GeoWave script a bit more aware of existing options only discoverable at run time through the installation of a plugin in a particular install.

PR #703, opened 2016-04-05
GEOWAVE-697: removed commandline completion

rfecher:
@spohnan this would best be reviewed by you, let me know if I missed something and let me know if it was too heavy-handed to remove completely and would be better commented out.  My assumption was that #635 would imply a fair amount of re-write so leaving traces of the old approach around is better left in our history than comments within the current baseline.


spohnan:
Tested packaging locally using the following steps and found an additional spec file update

```
# Set environment
export BUILD_ARGS="-Dfindbugs.skip=true -DskipFormat=true -DskipITs=true -DskipTests=true -Daccumulo.version=1.6.0-cdh5.1.4 -Dhadoop.version=2.6.0-cdh5.4.0 -Dgeotools.version=13.0 -Dgeoserver.version=2.7.0 -Daccumulo.api=1.6 -Dvendor.version=cdh5 -P cloudera"
export WORKSPACE="$(pwd)"

# Build
mvn install $BUILD_ARGS
deploy/packaging/rpm/admin-scripts/jenkins-build-geowave.sh

# Package
docker run --rm -e UID=$(id -u) -e BUILD_ARGS="$BUILD_ARGS" -v $(pwd):/usr/src/geowave geoint/geowave-centos6-rpm-build sh -c "cd /usr/src/geowave; deploy/packaging/docker/build-rpm.sh; chown -R \$UID:\$UID ." 
```

PR #705, opened 2016-04-05
ignoring files generated by eclipse's scala plugin

rfecher:


rfecher:
this one is so simple it is arguably not needing its own pr, just figured I'd separate it cleanly rather than have it comingled in some other pr

Issue #546, opened 2015-11-09
Improve parsing/handling of command line options


dcy2003:
The current approach, using Apache CLI is verbose from a code standpoint and rigid from a usage standpoint (e.g. expects options in a specific order).  Consider replacing Apache CLI with another library such as JCommander or Args4J for improved handling of command line options.

chrisbennight:
JCommander++

datasedai:
Related to #331  - Ensure that CLI errors are shown in a prettier format.

Issue #435, opened 2015-06-22
Add option for more logging from the CLI tool


rfecher:
Either '-v' for verbose and toggle more messaging on, or '--debug' which could automatically turn up the log level for log4j (without the user having to hassle with log4j properties)

In general, we need easier ways for users to get more insight into the CLI/ingest processing in cases where it is not matching their expected behavior.

datasedai:
#546 - Add option to programmatically up the log level for log4j.

Issue #416, opened 2015-06-04
CLIOperationDriver.run should return a boolean


rwgdrummer:
When false, the main line routine should exit with -1 or 1.

datasedai:
#546 Should make sure to implement this.

Issue #331, opened 2015-04-11
Cleanup stack trace for input validation in ingest framework


chrisbennight:
Comes from using the default input validator.  Technically it's fine, but it looks pretty bad.   Basically we should have a nicer message, not the Log4J FATALS + stack trace.

for example:

```
[bennight@cacti ~]$ geowave-ingest -localingest
11 Apr 16:49:03 FATAL [ingest.AccumuloCommandLineOptions] - Zookeeper URL not set
11 Apr 16:49:03 FATAL [ingest.AccumuloCommandLineOptions] - Accumulo instance ID not set
11 Apr 16:49:03 FATAL [ingest.AccumuloCommandLineOptions] - Accumulo user ID not set
11 Apr 16:49:03 FATAL [ingest.AccumuloCommandLineOptions] - Accumulo password not set
11 Apr 16:49:03 FATAL [ingest.AbstractCommandLineDriver] -
org.apache.commons.cli.ParseException: Required option is missing
        at mil.nga.giat.geowave.ingest.AccumuloCommandLineOptions.parseOptions(AccumuloCommandLineOptions.java:186)
        at mil.nga.giat.geowave.ingest.local.LocalFileIngestDriver.parseOptionsInternal(LocalFileIngestDriver.java:49)
        at mil.nga.giat.geowave.ingest.AbstractCommandLineDriver.applyArguments(AbstractCommandLineDriver.java:168)
        at mil.nga.giat.geowave.ingest.AbstractCommandLineDriver.run(AbstractCommandLineDriver.java:71)
        at mil.nga.giat.geowave.ingest.IngestMain.main(IngestMain.java:51)
usage: -localingest <options>

Options:
 -b,--base <arg>               Base input file or directory to crawl with
                               one of the supported ingest types
 -c,--clear                    Clear ALL data stored with the same prefix
                               as this namespace (optional; default is to
                               append data to the namespace if it exists)
 -dim,--dimensionality <arg>   The dimensionality type for the index,
                               either 'spatial' or 'spatial-temporal'
                               (optional; default is 'spatial')
 -h,--help                     Display help
 -i,--instance-id <arg>        The Accumulo instance ID
 -l,--list                     List the available ingest types
 -n,--namespace <arg>          The table namespace (optional; default is
                               no namespace)
 -p,--password <arg>           The password for the user
 -t,--types <arg>              Explicitly set the ingest type by name (or
                               multiple comma-delimited types), if not set
                               all available ingest types will be used
[...]

PR #723, opened 2016-04-20
GEOWAVE-546 - CLI Refactor

datasedai:
This update includes: #331 #416 #435 #546


analytics/api/src/main/java/mil/nga/giat/geowave/analytic/GeometryDataSetGenerator.java
```diff
@@ -490,92 +481,94 @@ private SimpleFeature createFeatureWithGeometry(
...
-						new double[] {
-							-90,
-							-90
-						},
-						new double[] {
-							90,
-							90
-						}));
-	}
+	// public static void main(
```
rfecher: do we use this anymore?  If not, can we just delete it?  (if so, we probably have a problem in the first place here, so I'll assume we don't)

rfecher:
@rwgdrummer @dcy2003 Do you guys *want* to review this?  Its pretty massive and I'd vote for getting it into 0.9.1 sooner than later so we can press forward.  I glanced through it and I really think thats all anyone is going to be able to do without setting aside a full day.  Let me know.

dcy2003:
I peaked through it the other day.  Nothing stood out, so I'm good with moving forward with it.

Issue #740, opened 2016-04-28
Remove dead code and reconnect stats option switch for Accumulo


rwgdrummer:

datasedai:
Talking to Eric earlier today, it appears that some merge issues as a result of the Accumulo IndexWriter update a month ago could have resulted in some missed merge conflicts...

PR #741, opened 2016-04-28
Issue #740. Reconnect  Accumulo options to turn off statistics

rwgdrummer:


datasedai:
Confused about this pull request.  It seems like it has a commit that I did in it, that's unrelated ...?

Issue #563, opened 2016-01-05
Avoid serialization of field IDs when combining attributes with shared visibility


dcy2003:
Currently the Accumulo optimization to flatten fields sharing a common visibility attribute into a single column qualifier serializes/deserializes the original field IDs alongside the values.  This approach increases the size required to store each feature and we also pay a performance penalty.

Investigate approaches to avoid the serialization/deserialization of field IDs while maintaining the ability to properly deconstruct the flattened value (preferably without knowledge of the data adapter) and support field level query filtering.

Ultimately this should result in a performance boost, but the tradeoff is additional complexity.  Goal is to find the best approach to achieve this performance boost while minimizing complexity.

PR #742, opened 2016-04-28
Accumulo Bitmask Optimization

dcy2003:
#563


extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/AttributeSubsettingIterator.java
```diff
@@ -0,0 +1,206 @@
...
+	 * @param adapterAssociatedWithFieldIds
+	 *            the adapter associated with the given fieldIds
+	 * @param fieldIds
+	 *            the desired subset of fieldIds
+	 * @param numericDimensions
+	 *            the numeric dimension fields
+	 */
+	public static void setFieldIds(
+			final IteratorSetting setting,
+			final DataAdapter<?> adapterAssociatedWithFieldIds,
```
datasedai: Stupid question ... if someone adds a new field in the future, it will always be added at the end, right?  Like, there's no chance that it could be added somewhere in the middle due to its type or name or ordering or something, and mess up the bitmasks?

Also, what happens to the bitmask AND operation earlier in this file, if the length of the bitmasks is different?

extensions/formats/stanag4676/format/src/main/java/mil/nga/giat/geowave/format/stanag4676/image/ImageChipDataAdapter.java
```diff
@@ -112,4 +113,43 @@ public void fromBinary(
...
+			final CommonIndexModel model,
+			final ByteArrayId fieldId ) {
+		int i = 0;
+		for (final NumericDimensionField<? extends CommonIndexValue> dimensionField : model.getDimensions()) {
+			if (fieldId.equals(dimensionField.getFieldId())) {
+				return i;
+			}
+			i++;
+		}
+		if (fieldId.equals(IMAGE_FIELD_ID)) {
```
datasedai: I know you have something like this everywhere ... what's going on exactly?  I see this 'fieldId.equals(CONSTANT)' in all the adapters but it doesn't make obvious sense to me immediately...

datasedai:
Did you re-test the other DistributableQueryFilters like the DateRange, or TextQuery? etc?  Do they still work?

datasedai:
...Reviewed ... Didn't see anything.

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/AttributeSubsettingIterator.java
datasedai: Another question.. How about MSB/LSB and Endian-ness?  If I have a client on windows and a accumulo iterator running on linux, does the bitmask implementation you're using translate everything to the same Endian-ness?

rfecher: Regarding the and operation on different length byte arrays, the operation produces the smallest length as the result - essentially if bytes are missing at the end they are assumed 0 or missing fields which should accurately reflect the behavior.

If someone were to add fields, correct, it would have to be at the end.

I don't think endian-ness comes into play here - its always bitwise operations on raw bytes without any translation to numbers.  As long as the bytes come across consistently in serialization/deserialization (which they will) after that point its just bitwise operations.

extensions/formats/stanag4676/format/src/main/java/mil/nga/giat/geowave/format/stanag4676/image/ImageChipDataAdapter.java
rfecher: this is a single field adapter, as is the rasterdataadapter...basically the field ID is constant and the adapter puts everything in a single field. So this is essentially saying "is the field the one expected value, if not return -1.  Because line 72 of this class in the encode() method just puts a single field with a constant field ID in the encoding, the only field that it should have is that constant field ID

Issue #733, opened 2016-04-27
Rework AttributeSubsettingIterator


dcy2003:
AttributeSubsettingIterator currently executes prior to QueryFilterIterator (if a subset of fields is provided).
It should be reworked to execute after QueryFilterIterator, so any filtering happens first, and attributes are only filtered from the results that pass all filters.  This should result in a performance boost for this use case.  Specifically, AttributeSubsettingIterator will need to be modified to operate on encoded "whole rows" rather than individual key-value pairs.

PR #753, opened 2016-05-03
Subset fields after filtering, rather than prior (performance improvement)

dcy2003:
Fix for #733


rfecher:
it'd be really nice to get some level of metrics on this comparing it to the previous implementation.  Is that something we can do quick and hold off merging this or should we merge this in and defer measuring the performance implication?

Issue #696, opened 2016-03-31
IndexWriter needs to be closed, flush isn't sufficient unless close is eventually called 


rfecher:
This may have to do with Statistics not being written on flush - needs further investigation.  I recommend using BasicKafkaIT to test this.

rwgdrummer:
This is not an issue any longer. Although, we should still add the test to BasicKafkaIT .

datasedai:
I have verified that this is still an issue.  Without .close(), the records are not written.

I also verified that a bug was introduced in #658.  An empty list of indexes is passed to toGeoWaveData, and as a result no data is written to GeoWave.

I am looking into both of these issues.

datasedai:
So, I think the reason this isn't working is that IndexCallbackList<T> is not Flushable.  The callback list is created in AccumuloDataStore.createWriter:
`			final IngestCallbackList<T> callbacksList = new IngestCallbackList<T>(
					callbacks);
`
But, the flush() method in AccumuloIndexWriter only flushes the callback if it is Flushable.  Adding "Flushable" interface to IngestCallbackList seems to have fixed the issue. Pull request soon.

PR #759, opened 2016-05-05
GEOWAVE-546 & GEOWAVE-696 Fixes / Updates

datasedai:
For GEOWAVE #546: If a kafka properties file isn't selected, then it won't apply command line options to the kafka config properties.  Changed it so that it will do it whether the config file is specified or not.

For GEOWAVE #696: Fixed IndexCallbackList to implement Flushable, which will then allow the IndexWriter.flush() method to flush properly.  Also, modified IT tearDown code to work in reverse order.  For Kafka IT, it was tearing down Accumulo (with zookeeper) before tearing down kafka, and the kafka client was in an infinite loop trying to re-connect to zookeeper at the end of the IT.

Also, modified the BasicKafkaIT to not wait randomly 60 seconds, but instead to wait until all the data has been ingested, and the consumer times out before continuing the integration test.  This will allow different systems with different performance specs to run the IT reliably.  Added it back to GeoWaveITSuite.

### DIFF
--- a/core/ingest/src/main/java/mil/nga/giat/geowave/core/ingest/kafka/IngestFromKafkaDriver.java
+++ b/core/ingest/src/main/java/mil/nga/giat/geowave/core/ingest/kafka/IngestFromKafkaDriver.java
@@ -7,8 +7,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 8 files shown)


core/ingest/src/main/java/mil/nga/giat/geowave/core/ingest/kafka/IngestFromKafkaDriver.java
```diff
@@ -155,20 +159,22 @@ private void configureAndLaunchPlugins(
...
 	}
 
-	private void launchTopicConsumer(
+	private Future<?> launchTopicConsumer(
 			final String formatPluginName,
 			final AvroFormatPlugin<?, ?> avroFormatPlugin,
 			final KafkaIngestRunData ingestRunData,
 			final List<String> queue )
 			throws Exception {
 		final ExecutorService executorService = Executors.newFixedThreadPool(queue.size());
```
datasedai: I just realized something here.. Is this creating a thread pool of X size for every single topic?  Like if I had 3 topics it would create 3 threadpools of 3 max threads each?  I think it would probably only actually create 3 threads but there might be overhead from the extra pools...

PR #767, opened 2016-05-10
Fixed thread safety issue for MemoryDataStore used in local ingest

rwgdrummer:


core/store/src/main/java/mil/nga/giat/geowave/core/store/memory/MemoryAdapterStore.java
```diff
@@ -3,6 +3,7 @@
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
```
datasedai: Can you apply this to the other stores as well?

PR #770, opened 2016-05-11
fixed concurrency issue in position map

rfecher:
with multi-threaded ingest previous implementation has a race condition that can result in non-deterministic exceptions of this form:
java.lang.IllegalArgumentException: value already present: 29
	at com.google.common.collect.HashBiMap.put(HashBiMap.java:246)
	at com.google.common.collect.HashBiMap.put(HashBiMap.java:223)
	at mil.nga.giat.geowave.adapter.vector.FeatureDataAdapter.initializePositionMaps(FeatureDataAdapter.java:710)

calling forcePut should at least circumvent this issue

### DIFF
--- a/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/FeatureDataAdapter.java
+++ b/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/FeatureDataAdapter.java
@@ -707,7 +707,7 @@ public class FeatureDataAdapter extends
 			final AttributeDescriptor ad = reprojectedType.getDescriptor(i);
 			final ByteArrayId currFieldId = new ByteArrayId(
 					ad.getLocalName());
-			fieldToPositionMap.put(
+			fieldToPositionMap.forcePut(
 					currFieldId,
 					i);
 		}


dcy2003:
Alternatively we can synchronize those bimaps to avoid calling forcePut, like this:

https://github.com/dcy2003/geowave/commit/5bb2da1ab2bbd7a366ef8c3d318b68567571b7f6

Thoughts?

rfecher:
I don't care much, although that synchronize is in the wrong place.  The race condition would still be introduced.  If you wanted to sync you'd have to wrap this in a synchronized block instead
```
synchronized(MUTEX){
 			if (fieldToPositionMap.isEmpty()) {
 				initializePositionMaps();
 			}
}
```
If its a synchronized map that just means that the calls to the methods are synchronized - if its empty and two threads are coincidentally initializing it there will be the same issue.

if we went with the above block, they'd both get you what you want and I think its a matter of which one is high performance - forcePut + no synchronization vs. put + synchronization

my intuition was the forcePut+ no synch, because the forcePut just happens once (or in rare circumstances multiple times in the case of the race condition), the sync would on the other hand apply to every call to get filed positions

dcy2003:
Gotcha, okay, you're approach is probably more performannt.  I looked at this quickly and misunderstood the race condition.  Thought it was related to map access.  Will merge once Travis passes.

PR #772, opened 2016-05-11
made zorder behave more like hilbert for bit precisions not divisible…

rfecher:
… by 8 - padded left-most bits rather than rightmost


rfecher:
this one is pretty meaningless considering we will prefer hilbert, but when using bits not divisible by 8 (let's say 10 bits would force you to represent it with 2 bytes so the last 6 bits would be 0 rather than the first 6 bits) we were padding 0's on the end rather than the beginning.  It makes more sense to pad to the beginning although both seem valid and would work, this behavior is consistent with hilbert SFC

PR #790, opened 2016-05-16
GEOWAVE-546 - Fix issue with Analytics

datasedai:
Added to fix issue where analytics properties do not set correctly from the command line.   Already added to 0.9.1 tag.


analytics/mapreduce/src/main/java/mil/nga/giat/geowave/analytic/mapreduce/operations/options/PropertyManagementConverter.java
```diff
@@ -1,150 +1,151 @@
...
+	private void handleEnum(
+			TranslationEntry entry,
+			ParameterEnum<?> enumVal ) {
+		Object value = entry.getParam().get(
+				entry.getObject());
+		if (value != null) {
+			((ParameterEnum<Object>) enumVal).getHelper().setValue(
+					properties,
+					enumVal.getHelper().getValue(
+							value.toString()));
```
datasedai: This is the only change in this file.

Issue #555, opened 2015-11-20
implement and expose new temporal binning strategies


rfecher:
currently temporal periodicities that are supported are "binning" by year, month, and day.

Other temporal units to support as binning strategies should include week (week of year), hour, and decade to provide more flexibility.

PR #801, opened 2016-05-24
added temporal units to support binning by decade, hour, and minute

dcy2003:
#555

### DIFF
--- a/core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/index/dimension/TemporalBinningStrategy.java
+++ b/core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/index/dimension/TemporalBinningStrategy.java
@@ -17,25 +17,34 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
  * This class is useful for establishing a consistent binning strategy using a
- * unit of time (currently day, month, or year). Each bin will then be defined
- * by the boundaries of that unit within the timezone given in the constructor.
- * So if the unit is year and the data spreads across 2011-2013, the bins will
- * be 2011, 2012, and 2013. The unit chosen should represent a much more
- * significant range than the average query range (at least 20x larger) for
- * efficiency purposes. So if the average query is for a 24 hour period, the
- * unit should not be a day, but could be perhaps a month or a year (depending
- * on the temporal extent of the dataset).
+ * unit of time. Each bin will then be defined by the boundaries of that unit
+ * within the timezone given in the constructor. So if the unit is year and the
+ * data spreads across 2011-2013, the bins will be 2011, 2012, and 2013. The
+ * unit chosen should represent a much more significant range than the average
+ * query range (at least 20x larger) for efficiency purposes. So if the average
+ * query is for a 24 hour period, the unit should not be a day, but could be
+ * perhaps a month or a year (depending on the temporal extent of the dataset).
  */
 public class TemporalBinningStrategy implements
 		BinningStrategy
 {
 	public static enum Unit {
+		MINUTE(
+				Calendar.MINUTE),
+		HOUR(
+				Calendar.HOUR_OF_DAY),
 		DAY(
 				Calendar.DAY_OF_MONTH),
+		WEEK(
+				Calendar.WEEK_OF_YEAR),
 		MONTH(
 				Calendar.MONTH),
 		YEAR(
-				Calendar.YEAR);
+				Calendar.YEAR),
+		DECADE(
+				-1);
+		// java.util.Calendar does not define a field number for decade
+		// use -1 since that value is unused
 
 		private final int calendarEnum;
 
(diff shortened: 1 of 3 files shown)


rfecher:
To expose it to the commandline tools you have t o add it as options to the periodicity in SpatialTemporalDimensionalityTypeProvider.  Also, if you could figure out a way to get "week" in there that would be awesome!

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/index/dimension/TemporalBinningStrategy.java
```diff
@@ -211,6 +237,17 @@ public int getFixedBinIdSize() {
...
+				return StringUtils.stringToBinary((Integer.toString(value.get(Calendar.YEAR)) + "_"
+						+ TWO_DIGIT_NUMBER.format(value.get(Calendar.MONTH)) + "_"
+						+ TWO_DIGIT_NUMBER.format(value.get(Calendar.DAY_OF_MONTH)) + "_" + TWO_DIGIT_NUMBER
+						.format(value.get(Calendar.HOUR_OF_DAY))));
+			case MINUTE:
+				return StringUtils.stringToBinary((Integer.toString(value.get(Calendar.YEAR)) + "_"
+						+ TWO_DIGIT_NUMBER.format(value.get(Calendar.MONTH)) + "_"
+						+ TWO_DIGIT_NUMBER.format(value.get(Calendar.DAY_OF_MONTH)) + "_"
+						+ TWO_DIGIT_NUMBER.format(value.get(Calendar.HOUR_OF_DAY)) + "_" + TWO_DIGIT_NUMBER
+						.format(value.get(Calendar.MINUTE))));
```
rwgdrummer: May be we should now consider compressing these a bit.

rfecher: perhaps a good idea - a req't for that compression is that for each periodicity/bin, the binary size must be fixed, but I don't think that should be a problem. preserving sort order actually shouldn't matter as we are just using the periodicity/bin as an enumeration

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/index/dimension/TemporalBinningStrategy.java
```diff
@@ -223,6 +260,20 @@ private Calendar getStartEpoch(
...
+				cal.set(
+						Calendar.MINUTE,
+						minute);
+			case HOUR:
+				final int hour = Integer.parseInt(str.substring(
+						11,
+						13));
+				cal.set(
+						Calendar.HOUR_OF_DAY,
+						hour);
```
rwgdrummer: Should we add DECADE?
Basically, it is  a variant to setting the YEAR.

rfecher: it seems that decade is essentially accounted for entirely within getBinSizeMillis() but I kinda agree with @rwgdrummer - a check for decade here would essentially ensure that the year is a multiple of 10 (rounding down if it is now).  Perhaps redundant but it is definitely consistent with what one would expect from getStartEpoch()

dcy2003: Good call.  Will do.

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/index/dimension/TemporalBinningStrategy.java
dcy2003: So what do we want to do here?  I can come up with an approach to compress these a bit, which will require me to rework the parsing logic within getStartEpoch(), and and adjust the fixed binId sizes.  Are we ok with this being a breaking change, or do we want to add complicated logic to handle both the new and old formats (a bit ugly)?

rfecher: I'd suggest you document an approach in an issue but don't worry about implementing it yet - tag the issue as "breaking changes"

dcy2003:
@rfecher not sure I follow your comment about exposing it in SpatialTemporalDimensionalityTypeProvider.  Isn't that handled via:
https://github.com/dcy2003/geowave/blob/GEOWAVE-555/core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/ingest/SpatialTemporalDimensionalityTypeProvider.java#L136-138
and parsed via:
https://github.com/dcy2003/geowave/blob/GEOWAVE-555/core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/ingest/SpatialTemporalDimensionalityTypeProvider.java#L214-232

core/geotime/src/main/java/mil/nga/giat/geowave/core/geotime/index/dimension/TemporalBinningStrategy.java
dcy2003: Opened #802

PR #804, opened 2016-05-25
Configure Travis to skip formatter to slightly speed up builds

dcy2003:


rfecher:
sorry @dcy2003, this unfortunately conflicts with master now

dcy2003:
@rfecher thanks for pointing that out.  Should be good to go now.

PR #825, opened 2016-06-14
GEOWAVE-809: row ID query support for a list of IDs

rfecher:
looks like there's some reformatting happening as well


andrewdmanning:
This should be OBE as these changes have been applied to the HBase refactor branch.

Issue #667, opened 2016-03-16
maintain statistics to be used as query hints


rfecher:
1 query hint would be what the full set of "visibility" possibilities are (or at least whether there is more than 1 visibility per adapter). If there is only 1 visibility throughout, then the whole row iterator is unnecessary.

another query hint is whether there is ever a duplicate per adapter-index pair (or perhaps the max number of duplicates)

another query hint is the filled tiers (some tiers may not have any data, the base case of point-timestamps only would just be a single tier). We only need to query the tiers that have data.

rwgdrummer:
Visibility Query Hint: Need a little more clarification on this hint.  Is the result of the whole row iterator where each column is separated into key/value pairs expected to change?
I was thinking the whole row iterator can be replaced to minimize the duplication column family and visibility for each column qualifier.

(2) Dedup: perhaps just a flag to turn duplicate checking off.  Without duplicate checking, we may see opportunity to replace the Row ID with something smaller at query time.

(3) Filled tiers: makes sense to me.

rfecher:
This is related to #48 but includes more statistics that could improve query

PR #830, opened 2016-06-17
GEOWAVE-667: added visibility stats and duplicate stats and used as q…

rfecher:
…uery hints

It is used on query now.  TBD is how we may want to use it for GeoWaveInputFormat (it'd probably be beneficial).  The stats would just need to be pulled into InputFormatAccumuloRangeQuery and passed along to AccumuloConstraintsQuery.  Also, the implementation just continued to primarily use patterns that had been established by IndexMetaDataSet, but it may be considered technical debt to try to clean up a little now that there's many stats being passed to AccumuloConstraintsQuery and now that there;s an ongoing set of stats that continue to be added to DataStoreStatisticsProvider (and ignored by GeoWaveBasicIT).  In general we probably need better hooks for users/developers to provide there own stats, and to subset or include more default stats.  

...So I think that this likely should become a new issue and I just (mostly) followed the pattern and added the appropriate stats given #667.


rfecher:
actually, I'm going to run benchmarks on this v. master so don't quite merge this in...yet.

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/WholeRowAggregationIterator.java
```diff
@@ -0,0 +1,183 @@
...
+
+	@Override
+	protected boolean filter(
+			final Text currentRow,
+			final List<Key> keys,
+			final List<Value> values ) {
+		if ((aggregationIterator != null) && (aggregationIterator.queryFilterIterator != null)
+				&& aggregationIterator.queryFilterIterator.isSet()) {
+			final PersistentDataset<CommonIndexValue> commonData = new PersistentDataset<CommonIndexValue>();
+			final List<AccumuloFieldInfo> unknownData = new ArrayList<AccumuloFieldInfo>();
```
rwgdrummer: This is a bit odd and deserves an explanation.
The check applies to the aggregationIterator and its queryFilterIterator.
The code below uses 'this' class instance queryFilterIterator

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/AggregationIterator.java
```diff
@@ -46,131 +48,195 @@
...
+									// serialization
+									// of
+									// the
+									// strategy
+									// in
+									// the
+									// options
+									// of
+									// this
+									// iterator
```
rwgdrummer: comment format madness

andrewdmanning: This comment seems to be too far indented for the formatter.

extensions/datastores/accumulo/src/main/java/mil/nga/giat/geowave/datastore/accumulo/query/WholeRowAggregationIterator.java
rfecher: good catch - we must not test aggregations where wholerowiterator is needed (aggregations on mixed visibilities) because it seems to me the this.queryFilterIterator below should be aggregationIterator.queryFilterIterator...this got messy trying to re-use as much code as possible AND have as efficient performance as possible between the variations of query, aggregations, and wholerow.

PR #837, opened 2016-06-22
Refactor and cleanup of initial HBase implementation.

andrewdmanning:


core/store/src/main/java/mil/nga/giat/geowave/core/store/util/EntryIteratorWrapper.java
```diff
@@ -71,18 +55,23 @@ private void findNext() {
...
-				row.getKey(),
-				row.getValue(),
-				adapterStore,
-				clientFilter,
-				index,
-				scanCallback);
-	}
+			final PrimaryIndex index );
+
+	// private T decodeRow(
```
rfecher: is there a reason these commented lines were left here?

andrewdmanning: Nope, oversight.

PR #841, opened 2016-06-29
Initial create for geoserver cli

blastarr:


dcy2003:
Reviewed.  Looks good.  Does it make sense to add any tests for this functionality?  (not necessarily right now, possibly just add a new issue to the backlog?)

rfecher:
@blastarr does have some tests that he was firing off through shell scripts, I could be wrong but I don't think it'd be too much trouble to make it an integration test...yeah, a separate issue for the backlog

PR #844, opened 2016-07-01
Geowave 833

mawhitby:
Made some small changes to the gdelt-example for use with the new geowave quickstart guide.

### DIFF
--- a/analytics/mapreduce/src/main/java/mil/nga/giat/geowave/analytic/mapreduce/operations/KdeCommand.java
+++ b/analytics/mapreduce/src/main/java/mil/nga/giat/geowave/analytic/mapreduce/operations/KdeCommand.java
@@ -1,6 +1,7 @@
 package mil.nga.giat.geowave.analytic.mapreduce.operations;
 
 import java.io.File;
+import java.io.FilenameFilter;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


analytics/mapreduce/src/main/java/mil/nga/giat/geowave/analytic/mapreduce/operations/KdeCommand.java
```diff
@@ -62,6 +64,29 @@ public KDEJobRunner createRunner(
...
+			if (configFiles != null && configFiles.length > 0) {
+				final String firstFile = String.format(
+						"%s%s%s",
+						defaultPath.getAbsolutePath(),
+						File.separator,
+						configFiles[0]);
+				configFile = new File(firstFile);
+			}
+			
+		}
```
datasedai: Just curious why this code is here ... but the code in ConfigOptions looks fairly similar?  I haven't dug deep into the code so they may be doing different things, but I would hope that we would just pass the better config file as ConfigOptions.PROPERTIES_FILE_CONTEXT instead of putting this snippet of code everywhere in the baseline?

mawhitby: Sorry about that. I created the pull request from the wrong branch. Still looking into this issue. As you can tell from the comment I meant to submit a pull request for the gdelt-example script changes.

PR #867, opened 2016-07-18
Twitter ingest format v0.9.0

vpipkt:
For v0.9.0 an extension to provide ingest of twitter API tweets. Each tweet is expected to well formed json with no newlines in a (possibly compressed) text file.


vpipkt:
Although the CI says it builds successfully, the twitter plugin was not included in the build. It must be refactored.

rfecher:
refactored to support latest on master, pulled in #899

PR #881, opened 2016-08-23
adding fuctionality for quickstart guide use

mawhitby:


mawhitby:
we will need to up our head size allocation on Travis builds due to the larger documentation size

PR #897, opened 2016-09-13
fixed an oversight in index metadata statistics

rfecher:


core/index/src/main/java/mil/nga/giat/geowave/core/index/CompoundIndexStrategy.java
```diff
@@ -544,9 +544,6 @@ public static ByteArrayId extractByteArrayId(
 		}
 
 		final byte[] bytes2 = new byte[id.getBytes().length - id1Length - 4];
-		if (id1Length < 0) {
-			System.err.println("Crap");
```
dcy2003: 💩  = 😂

PR #919, opened 2016-10-19
Set up travis to run unit- or integration-tests only

blastarr:
Just a re-imagining of our Travis build. Now we can run a unit-test-only build, followed by the full suite of IT env's, but without unit tests.  
  
The unit tests run in under 20 minutes, and the IT's without them run in under 40. Because of the faster execution time, we can use the "trusty" Travis build (8GB, 50-min limit).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,20 @@
-sudo: false
+sudo: required
+dist: trusty
 language: java
 jdk:
   - oraclejdk7
 env:
   matrix:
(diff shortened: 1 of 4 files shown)


dcy2003:
:ok_hand: :thumbsup: :clap:

blastarr:
The unit-test-only is running out of memory - some tweaks to do. Also, want to pull the dev-resources from our repo on S3 instead of building every time.

blastarr:
OK, memory is increased, and the S3 pull is working.

PR #975, opened 2016-12-19
adding previous versions of documentation

mawhitby:


mawhitby:
Adding the previous-versions directory so that we have a place to keep older versions of the documentation. Our build scripts will update this directory automatically.

PR #1008, opened 2017-01-24
cassandra passing a few basic ITs

rfecher:


core/store/src/main/java/mil/nga/giat/geowave/core/store/util/DataStoreUtils.java
```diff
@@ -552,16 +607,11 @@ else if ((vis2 == null) || (vis2.length == 0)) {
...
-		final int totalLength = idBuf.array().length;
-		final byte[] dataId = new byte[idBuf.getInt()];
-		idBuf.get(dataId);
-
-		final byte[] adapterId = new byte[totalLength - dataId.length - 4];
-		idBuf.get(adapterId);
+			final ScanCallback<T, R> scanCallback ) {
+		final byte[] dataId = row.getDataId();
+		final byte[] adapterId = row.getAdapterId();
 		final GeowaveRowId rowId = new GeowaveRowId(
```
blastarr: The params for this rowId appear to be in the wrong order. It's passing index, dataId, adapterId, but the constructor is defined as dataId, adapterId, index.

rfecher: I don't recall exactly why I changed the constructor, probably to be more consistent with the order they are combined in a row key, but the constructor changed to match this change.

PR #1014, opened 2017-02-03
Move InstallGdal

mak438:
Move InstallGdal to extensions/adapters/raster and update dependencies; unit test for Landsat8IngestCommandTest to use InstallGdal.

### DIFF
--- a/extensions/adapters/raster/pom.xml
+++ b/extensions/adapters/raster/pom.xml
@@ -93,5 +93,39 @@
 			<version>4.11</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.codehaus.plexus</groupId>
+			<artifactId>plexus-archiver</artifactId>
+			<version>2.2</version>
+		</dependency>
+                <dependency>
+                        <groupId>org.apache.kafka</groupId>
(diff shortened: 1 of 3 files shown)


extensions/adapters/raster/pom.xml
```diff
@@ -93,5 +93,39 @@
 			<version>4.11</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
mak438: Dependencies needed for InstallGdal.

extensions/adapters/raster/src/main/java/mil/gma/giat/geowave/gdal/InstallGdal.java
```diff
@@ -63,15 +117,15 @@ public static void main(
 			}
 		}
 		if (file.endsWith("zip")) {
-			TestUtils.unZipFile(
+			unZipFile(
```
mak438: Can't create a dependency on TestUtils from here. Otherwise, the dependency becomes cyclic.

rfecher:
make sure GDAL is on the PATH and on LD_LIBRARY_PATH

Issue #999, opened 2017-01-20
Add Unit tests to the package mil.nga.giat.geowave.format.landsat8


rfecher:
There are currently no unit tests of any of the public methods in the java classes inside the mil.nga.giat.geowave.format.landsat8 package.  These are commands exposed to the GeoWave command line tool that interact with the publicly available LandSat 8 dataset to make it easier for users to understand what’s available and ingest the data into GeoWave.

mak438:
Should the unit tests be for the Command classes, or do the Iterator classes in this package require tests for their close(), next(), hasNext() methods as well?

rfecher:
The *Runner classes do most of the work so I'd think they'd be prime candidates for tests, but also SceneFeatureIterator and WRS2GeometryStore are complex enough that they should have some testing.  The SceneFeatureIterator abstracts the concept of downloading the list of scenes from a public URL and filtering through the list based on user specified criteria, and the WRS2GeometryStore abstracts the concept of downloading the shapefile that contains all of the row/path pairs that represent the footprints that Landsat8 is collecting globally and iterates through the geometries based on some user-specified filter criteria.

mak438:
Hi @rfecher. Thanks for the info. I began working on the tests for Landsat8AnalyzeCommand and Landsat8DownloadCommand (the *Runner classes don't have a public execute method and so I chose to look at the *Command classes instead).

I've noticed that both of these perform a large download and extract some files as part of their operation. I have some questions regarding the scope of these tests to make them most useful:

For the analyze command, I set up a test that calls Landsat8AnalyzeCommand#execute(), which downloads the files from the public URL. The test then verifies that certain values were printed to stdout ("Total Scenes: ", "Date Range: ", etc). Is there a way to mock the data source where this is fetched, perhaps by having a URL flag in Landsat8BasicCommandLineOptions? This would make the test go faster (since it doesn't need to download as much) and would let the test check that the computed totals are correct (I'm cautious to add that just yet in case data changes from http://landsat-pds.s3.amazonaws.com/ cause tests to fail).

The download command also loads some pretty large files (at least 150MB worth of images, could be more). Could that be mocked to load from a different source, perhaps something locally in a test folder, which would speed up that test as well? I haven't gotten this test completely working yet, so I'm not sure what else this entails.

In other words, is there a way (and would it be more useful) to have the tests be independent of the landsat data source, and simply validate that whatever's loading the data is working correctly?

Thanks in advance.

rfecher:
[...]

PR #1019, opened 2017-02-10
Geowave 999

mak438:
#999


mak438:
Apparently one of the new tests segfaulted during the CI build. Looking into that.

extensions/cli/landsat8/src/main/java/mil/nga/giat/geowave/format/landsat8/Landsat8IngestCommand.java
```diff
@@ -21,7 +21,7 @@
 {
 
 	@Parameter(description = "<storename> <comma delimited index/group list>")
-	private final List<String> parameters = new ArrayList<String>();
+	protected final List<String> parameters = new ArrayList<String>();
```
mak438: Note to self: undo this.

extensions/cli/landsat8/src/main/java/mil/nga/giat/geowave/format/landsat8/AnalyzeRunner.java
```diff
@@ -181,6 +181,10 @@ private void printSceneInfo() {
 					final String bandId = entry.getKey();
 					final SimpleFeature feature = entry.getValue();
 					if (first) {
+						if (System.out == null) throw new RuntimeException(
```
mak438: Note to self: undo this.

rfecher:
resubmitted as PR #1036

I tweaked the filter to add ".. and sizeMB < 1" just to make sure we're getting small files (considering there was no time constraint, as imagery is collected in the future there are no guarantees of the size without this.

I also add calls to set the scale which essentially scales the image first before ingest, and it therefore finished quickly.

Lastly, just squashed commit history down

PR #1023, opened 2017-02-12
Adds tests verifying correct path is requested on most get* and add* methods for GeoServerRestClient.java

mrkev:
This also adds Mockito 1.9.5 as an external dependency in `test/pom.xml`


rfecher:
I fixed a few issues and resubmitted as PR #1033  - this test was lumped into the integration tests (all tests in geowave-test are ITs) rather than a unit test all other projects have a src/test/java folder for unit tests and withiout being added to GeoWaveITSuite it wouldn't have been run (resulting in less coverage).

I formatted it, cleaned a few things up and added it instead as a unit test.

PR #1031, opened 2017-02-12
Commit changes for Accumulo SplitsProvider tests as one commit

akash-peri:
Tests populateIntermediateSplits, wrapRange and unwrapRange


rfecher:
this fail our `mvn verify -P compatibility` test that ensures backwards compatibility to accumulo 1.6.  Unfortunately, we need to use maven munge plugin to perform conditional compilation based on an accumulo.api flag.  I added the necessary changes to compile on accumulo 1.6 and resubmitted PR #1035

PR #1044, opened 2017-02-27
[1026] Adding support for asynchronous query in DynamoDBQuery

anicr7:
Couple of open questions -

1.  What should be the value of "MAX_ASYNC_QUERY_RESULTS"?

2. Is there anyway to prevent using locks+signals directly in Async paginated query?


rfecher:
It looks really good to me, a couple considerations that I have is:
- it seems like the constructor could be called many times and it always kicks off an async query initially (basically as I read it the constraint on MAX_ASYNC_QUERY_RESULTS is not used on constructing the paginated query, seemingly leading to the ability to exceed the global max).  This isn't a particularly big issue, but couldn't we simply compare total async queries with max before calling queryasync() in the constructor?
- onError, it triggers noResult.signal() on line 122 but it seems if a thread was waiting in the nextIterator() method for results, it would get stuck by line 80 (in the while loop).  My concern is in the situation that we get an error from dynamodb could we get stuck in a deadlock when trying to use the iterator?  My thought would be onError we would need to set something that indicated the nextIterator() return null (maybe add `asyncQueryResults.add(null);` on line 121 but then check for null on line 91
- a small nit, `asyncQueryResults.isEmpty() == true` is simpler as just `asyncQueryResults.isEmpty()`

And in answer to your questions:
1) I'd default it to something much smaller, maybe 100, but I think it should be configurable.  We can make an option for something like `queryThreads` and even set the boolean for not using the async queries if the option is set to <= 1.
2) I think the nextIterator method *has* to block if there is an outstanding query request that it needs to wait for.  So there is no way to get around some locking and triggering mechanism at least in that instance, and the use of await+signal seems like reasonable approach for this.  I haven't quite gotten a full understanding of each and every use of lock/unlock and synchronized. I think that you are using these for thread safety and at a glance they look like they are needed (I typically just use synchronized blocks throughout which I believe is implicitly the exact same behavior as your explicit use of the lock/unlock try-finally blocks).  If you feel synchronized is cleaner you should be able to get the same behavior.

Also, I believe you mentioned tests.  AsyncPaginatedQueryTest would probably be a good set of unit tests to have given the complexity.

anicr7:
@rfecher,

I have made changes according to your comments above. Here is a brief summary of the changes

1. Use a intrinsic sync block instead of specific re-entrant locks using try and finally.
2. Use linked list instead of array deque to allow inserting null locks. This prevents wait forevers when we have exceptions in async queries
3. Make sure we don't fire queries immediately on construction if there are already too many async queries piled up.
4. Reduce the max async query limit to 100. This can be made configurable through metadata in the future

Let me know if there are any other changes that are needed

rfecher:
I think there was some caching issue with travis - I resubmitted the PR from a different branch (fresh cache) and it passed.  Weird!  (the resubmitted PR is approved, and I'm closing this)

PR #1048, opened 2017-03-07
HBase Visibility

blastarr:


extensions/datastores/hbase/src/main/java/mil/nga/giat/geowave/datastore/hbase/operations/BasicHBaseOperations.java
```diff
@@ -38,6 +45,7 @@
 	private final Connection conn;
 	private final String tableNamespace;
 	private final boolean schemaUpdateEnabled;
+	private final HashMap<String, List<String>> coprocessorCache = new HashMap<String, List<String>>();
```
rfecher: perhaps outside of the scope of this issue, but food for thought it may be nice to have something like the Cassandra implementation of KeyspaceStatePool that basically caches stateful data so that it can be used across instances of the Operations.  In other words, if the lifecycle of these operations objects is short-lived the cache could become less useful, but definitely not something that needs immediate attention.

test/src/main/java/mil/nga/giat/geowave/test/HBaseStoreTestEnvironment.java
```diff
@@ -101,6 +160,43 @@ public void setup() {
...
+								conn.getConfiguration(),
+								auths,
+								User.getCurrent().getName());
+
+						// Verify hfile version
+						String hfileVersionStr = conn.getAdmin().getConfiguration().get(
+								"hfile.format.version");
+						assertTrue(
+								"HFile version is incorrect",
+								hfileVersionStr.equals("3"));
```
rfecher: this is less of a problem because we control the test environment, but still, maybe it'd be more flexible to make this >= 3?

PR #1051, opened 2017-03-14
Ap casanamo mapreduce

akash-peri:
This adds support for MapReduce to DynamoDB
Note: This gets one of the two tests in BasicMapReduceIT working; the other one fails due to an issue with adapter ids that was fixed for HBase; a similar fix needs to be applied here


akash-peri:
Travis failed; need to investigate

rfecher:
yes, I'd suggest disabling DynamoDB support for BasicMapReduceIT until we get that other test straightened out.  Also, it looks like it failed before it got to that point based on a findbugs issue.  Make sure you can run `mvn verify` in extensions/datastores/dynamodb (it'll probably fail right now and show the details of the findbugs issue).

rfecher:
is this superceded by PR #1057?  Can we close this one?

akash-peri:
Yep, we can.

Issue #1012, opened 2017-02-01
Retain Spatial Predicates


bmendell:
CQLQuery.createOptimalQuery() is often used to attempt to re-use SpatialQuery (faster) when appropriate 
http://github.com/ngageoint/geowave/blob/master/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/query/cql/CQLQuery.java#L63 .  Within this code path, we need to retain the spatial predicate from the CQL expression or Filter object that is passed to createOptimalQuery().  The proper place to track which predicate is used in the expression is within the ExtractGeometryFilterVisitor http://github.com/ngageoint/geowave/blob/master/extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/plugin/ExtractGeometryFilterVisitor.java .

PR #1055, opened 2017-03-22
Added capability to retain geometric predicate associated with query geometry extracted from CQL expression

ashish217:
#1012


ashish217:
FYI: There were couple of files (not related to this issue) got pulled in this change due to them being not formatted correctly on the master.

ashish217:
When I re-run the build on Travis, everything passed without any failures.

PR #1066, opened 2017-04-07
Enabled Authorization Provider for the raster data adapter plugin

jwomeara:
Looks good to me!


jwomeara:
I'm not sure why this build is failing.  The docs are failing to build apparently because it can't find my geowave-adapter-auth artifact, but it built just fine in my original branch...  I tried building locally and that worked for me as well.  Any suggestions?

jwomeara:
Good to go?

rfecher:
the PR build was cancelled...restarting

jwomeara:
What?  The pull request build completed: https://travis-ci.org/ngageoint/geowave/builds/220623394

It's the branch build which was canceled.

rfecher:
yeah, ok the branch was cancelled then, why are the dependencies marked provided again?

jwomeara:
Because they conflict with the versions that are already included within geowave-adapter-vector and geowave-adapter-raster.  They are pulled in as transitive dependencies of geoserver.  We could go through and use dependency management to avoid that issue, but I figured it would be easier to just let geoserver decide which version we use.

rfecher:
I'm confused because they look to match the versions from geoserver 2.10.0

Also, `<scope>${geotools.scope}</scope>` would be better I think, and yeah, using the dependencymanagement in the parent pom would also be better.  The reason for geotools.scope is that these dependencies are packaged for other uses than just geoserver, and we want to make sure we don't get classnotfoundexceptions other places

extensions/adapters/auth/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+    </parent>
+    <artifactId>geowave-adapter-auth</artifactId>
+    <name>GeoWave Adapter Auth</name>
+    <description>Authorization functionality for GeoWave Data Adapters</description>
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-core</artifactId>
+            <version>4.0.4.RELEASE</version>
+            <scope>provided</scope>
```
jwomeara: You aren't looking at the latest commit.  Try this: https://github.com/ngageoint/geowave/pull/1066/commits/83f11a062799f26ff77518c6329c78838204ee3e#diff-c3f3a0b58ea56b7b1532923859181c72R15

jwomeara:
:clap:

Issue #1112, opened 2017-06-02
Decompose complex CQL query into GeoWave SpatialQuery correctly


ashish217:
example query:
BBOX("pickup_location",-73.575,40.3,-74.375,41.1) AND BBOX("dropoff_location",-73.852,40.739,-73.835,40.755) AND "pickup_datetime" <= '2009-09-14T11:59:59 EDT' AND "pickup_datetime" >= '2009-08-31T12:00:00 EDT' AND "dropoff_datetime" <= '2009-09-14T11:59:59 EDT' AND "dropoff_datetime" >= '2009-08-31T12:00:00 EDT'

ashish217:
#1115

PR #1115, opened 2017-06-08
Extract only the specified geometry attribute

ashish217:
Fixed an issue to extract out correct geometry constraints and predicate when you have multiple geometric attributes inside a CQL expression


ashish217:
#1112

PR #1153, opened 2017-08-08
HOTFIX: remove transient from geom serialization and fix temp view

blastarr:


blastarr:
Sorry - one last commit w/ formatting & error message cleanup

PR #1173, opened 2017-09-01
Proposing some pom dependency updates to newer versions which have al…

mcarrier7:
…ready been approved in the ATO process


mcarrier7:
Looking at https://github.com/GoogleCloudPlatform/cloud-bigtable-client/issues/1475, it looks like netty-tcnative-boringssl-static 2.0.0.Final version and later have issues with hadoop, so leaving at original 1.1.33.Fork19 version until this can be resolved

PR #1231, opened 2017-11-14
Fixed geoserver config command for rest calls and added a response

emacthecav:


extensions/cli/geoserver/src/main/java/mil/nga/giat/geowave/cli/geoserver/ConfigGeoServerCommand.java
```diff
@@ -292,11 +252,24 @@ public Void computeResults(
...
 		}
 
+		// save properties from ssl configurations
+		sslConfigOptions.saveProperties(existingProps);
+
 		// Write properties file
 		ConfigOptions.writeProperties(
-				propFile,
-				existingProps);
-		GeoServerRestClient.invalidateInstance();
```
rfecher: this wa important

PR #1232, opened 2017-11-14
HP Fortify false positives

mawhitby:


core/ingest/src/main/java/mil/nga/giat/geowave/core/ingest/local/AbstractLocalFileDriver.java
```diff
@@ -163,6 +169,8 @@ protected void processInput(
 				return;
 			}
 
+			// HP Fortify "Null Dereference" false positive
+			// Null checking is done in the next line
 			String s3EndpointUrl = configProperties.getProperty(ConfigAWSCommand.AWS_S3_ENDPOINT_URL);
```
rfecher: its talking about configProperties, this one's a true-positive (line 151 configProperties is set to null and is only set to non-null within a conditional)

core/ingest/src/main/java/mil/nga/giat/geowave/core/ingest/s3/S3URLConnection.java
```diff
@@ -67,6 +67,8 @@ private ClientConfiguration buildClientConfig() {
 				PROP_S3_HANDLER_SIGNER_OVERRIDE,
 				null);
 
+		// HP Fortify "Portability Flaw" false positive
+		// This is purposefully kept dynamic so help solve portability problems.
 		final ClientConfiguration clientConfig = new ClientConfiguration()
```
rfecher: lets just delete toLowerCase on line 65 and change equals on line 73 to equalsIgnoreCase()

PR #1233, opened 2017-11-14
Small fix for the setls cli command

JWileczek:
Was missing the call to  actually execute the command and print results.


rfecher:
covered by PR #1231

PR #1260, opened 2017-12-02
Provided status endpoint for REST operations and return data in JSON format

richard3d:
This work includes updates to ServiceEnabledCommand (and classes that derive from it) to provide a hint to services on whether the operation should be run sync/async.


core/ingest/src/main/java/mil/nga/giat/geowave/core/ingest/operations/ListPluginsCommand.java
```diff
@@ -42,6 +42,11 @@ public void execute(
 					string);
 		}
 	}
+	
+	@Override
+	public boolean runAsync() {
```
rfecher: this shouldn't be async

PR #1269, opened 2017-12-18
Initial gRPC work for vector query service

richard3d:
This work includes a package for shading the parts of gRPC/protobuf 3 that conflict with imports from 2.5


services/grpc/src/main/java/mil/nga/giat/geowave/service/grpc/GeoWaveGrpcServer.java
```diff
@@ -0,0 +1,333 @@
...
+				SpatialQueryParameters request,
+				StreamObserver<Feature> responseObserver ) {
+
+			final String storeName = request.getBaseParams().getStoreName();
+			final StoreLoader storeLoader = new StoreLoader(
+					storeName);
+
+			final ByteArrayId adapterId = new ByteArrayId(
+					request.getBaseParams().getAdapterId().toByteArray());
+			final ByteArrayId indexId = new ByteArrayId(
```
rfecher: indexID and adapter ID are typically optional params on a query

geowave-grpc-protobuf-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>mil.nga.giat</groupId>
+  <artifactId>geowave-grpc-protobuf-shaded</artifactId>
+  <version>0.0.1-SNAPSHOT</version>
```
rfecher: also, I think we should version this with geowave versions (just make it use the geowave parent)

services/grpc/src/main/java/mil/nga/giat/geowave/service/grpc/GeoWaveGrpcServer.java
richard3d: Leaving these parameters as discussed. They are indeed optional as well when invoking the gRPC call.

services/grpc/src/main/java/mil/nga/giat/geowave/service/grpc/GeoWaveGrpcServer.java
```diff
@@ -0,0 +1,349 @@
...
+			try {
+				queryGeom = new WKTReader(
+						JTSFactoryFinder.getGeometryFactory()).read(geomDefinition);
+
+				temporalConstraints = new mil.nga.giat.geowave.core.geotime.store.query.TemporalConstraints(
+						temporalRanges,
+						"ignored"); // the name is not used in this case
+			}
+			catch (final Exception e) {
+				LOGGER.error("Exception encountered:", e.getStackTrace().toString());
```
rfecher: I had meant something like  `LOGGER.error("Exception encountered reading geometry", e);`

geowave-grpc-protobuf-shaded/src/main/java/mil/nga/giat/geowave/service/grpc/protobuf/util/ByteStringContainer.java
```diff
@@ -0,0 +1,23 @@
+package mil.nga.giat.geowave.service.grpc.protobuf.util;
```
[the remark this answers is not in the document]
richard3d: It is used by gRPC clients when making a call through the stub. Otherwise, we have another reliance on protobuf's ByteString (this is part of the shade/relocation strategy). In the future I believe we should actually look at using a shaded version of hbase Kent and I came across. It shades it's usage of protobuf 2.5 - that way anything moving forward can just include protobuf 3 no problem. At the moment we are kinda doing things backward (everything using protobuf 3 needs to shade at the moment).

services/grpc/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+			<artifactId>geowave-deploy</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>mil.nga.giat</groupId>
+			<artifactId>geowave-grpc-protobuf-shaded</artifactId>
+			<version>0.0.1-SNAPSHOT</version>
+		</dependency>
+		<dependency>
+				<groupId>com.google.guava</groupId>
```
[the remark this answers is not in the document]
richard3d: I agree but we need the newer version for gRPC, so far it works ok with hbase.

services/grpc/src/main/java/mil/nga/giat/geowave/service/grpc/GeoWaveGrpcServer.java
```diff
@@ -0,0 +1,333 @@
...
+			Geometry queryGeom = null;
+			try {
+				queryGeom = new WKTReader(
+						JTSFactoryFinder.getGeometryFactory()).read(geomDefinition);
+			}
+			catch (final Exception e) {
+				LOGGER.error("Exception encountered:", e.getStackTrace().toString());
+			}
+
+			final QueryOptions options = new QueryOptions(
```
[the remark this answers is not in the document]
richard3d: Yes if the string is empty I set the variable to null actually that way the type is explicit instead of accidentally passing in empty string like we discussed.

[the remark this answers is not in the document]
rfecher: oh, missed it on line 205-208

services/grpc/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+	<dependencies>
+		<dependency>
+			<groupId>mil.nga.giat</groupId>
+			<artifactId>geowave-deploy</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>mil.nga.giat</groupId>
+			<artifactId>geowave-grpc-protobuf-shaded</artifactId>
+			<version>0.0.1-SNAPSHOT</version>
```
rfecher: shouldn't this be `${project.version}` now that the shaded artifact is part of the same versioning as the rest of geowave?

PR #1296, opened 2018-03-26
adding spark ingest and custom crs to docs

mawhitby:


docs/content/walkthrough-raster/001-hw-quickstart-guide-raster-demo.adoc
```diff
@@ -100,9 +100,14 @@ Add a spatial index.
 
 [source, bash]
 ----
-geowave config addindex -t spatial spatial
+geowave config addindex -t spatial spatial --crs 3857
 ----
 
+[NOTE]
+====
+We are setting a custom CRS here as an example. If you don't set this GeoWave defaults to EPSG:4326.
```
rfecher: can we use the term Coordinate Refeerence System here (it is the quickstart guide after all)

docs/content/userguide/045-ingestoverview.adoc
```diff
@@ -9,6 +9,8 @@
 
 In addition to the raw data to ingest, the ingest process requires an adapter to translate the native data into a format that can be persisted into the data store. Also, the ingest process requires an Index that is a definition of all the configured parameters that define how data is translated to Row IDs (how it is indexed). It also includes what common fields need to be maintained within the table to be used by fine-grained and secondary filters.
 
+There are various ways to ingest data into a GeoWave store. The standard localToGW command is used to ingest files from locally or from an AWS s3 bucket into GeoWave in a single threaded fashion. For a distributed ingest (recommended for larger datasets) the sparkToGW and mrToGW commands can be used. Ingests can also be performed from directly from HDFS or utilizing Kafka.
```
rfecher: "ingest files from locally" should change to "ingest files from a local file system" , we can capitalize "S3", and 
 "performed from directly from HDFS" remove the extra "from"
change "in a single threaded fashion" to "in a single process." This is just clearer and more correct because localToGw has an option --numThreads which runs it multi-threaded.
also `localToGW` can ingest from HDFS the same as S3, `sparkToGW`.  @srinivasreddyv2 can clarify if needed.

PR #1299, opened 2018-03-27
HP fortify Fals Positives

mawhitby:


analytics/spark/src/main/java/mil/nga/giat/geowave/analytic/spark/spatial/TieredSpatialJoin.java
```diff
@@ -393,6 +393,8 @@ private void addMatches(
 						BasicNumericDataset convertedBounds = new BasicNumericDataset(
 								boundsRange);
 						List<ByteArrayId> insertIds = targetStrategy.getInsertionIds(convertedBounds);
```
rfecher: If it's referring to targetStrategy, it can be null if target tier ID isn't going in the above loop.

PR #1302, opened 2018-03-28
Refactor SparkConf and SparkSession within GeoWave

JWileczek:
Creates a class to help simplify the creation of SparkConf that will work with GeoWave. 
Includes some formatting changes not caught in last master build.
Can add onto this class in the future i.e. change how default options are loaded if we need different per environment, and also better ensures the user will have Kryo serializer set.


JWileczek:
Closing PR in favor of newly posted one with cleaner history.

PR #1315, opened 2018-04-04
All GRPC services implemented

richard3d:
This includes an example/template gRPC server and also automation for the generation of protobuf files from service enabled commands.


rfecher:
superseded by PR #1385

PR #1342, opened 2018-06-07
Add options to gpx ingest to simplify geometry

JWileczek:
and exclude tracks that are too complex or cover too much area. 
Simplfication still needs to be moved to generalized option provider within AbstractSimpleFeatureIngest.


extensions/formats/gpx/src/main/java/mil/nga/giat/geowave/format/gpx/GpxIngestFormat.java
```diff
@@ -22,10 +26,14 @@
 public class GpxIngestFormat extends
 		AbstractSimpleFeatureIngestFormat<GpxTrack>
 {
+	private final GeometrySimpOptionProvider simplifyOptProvider = new GeometrySimpOptionProvider();
```
rfecher: I thought you intended to put this in `AbstractSimpleFeatureIngestFormat` so it was more broadly available to other formats?

PR #1345, opened 2018-06-13
Add options to gpx ingest to simplify geometry, and exclude tracks that are too complex or cover too much area.

JWileczek:


extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/ingest/AbstractSimpleFeatureIngestPlugin.java
```diff
@@ -66,14 +67,21 @@ public void setTypeNameProvider(
...
+			final GeometrySimpOptionProvider geometryProvider ) {
+		this.simpOptionProvider = geometryProvider;
+	}
+
 	@Override
 	public byte[] toBinary() {
 		final byte[] filterBinary = filterOptionProvider.toBinary();
 		final byte[] typeNameBinary = typeNameProvider.toBinary();
-		final ByteBuffer buf = ByteBuffer.allocate(filterBinary.length + typeNameBinary.length + 4);
+		final byte[] simpBinary = simpOptionProvider.toBinary();
```
rfecher: fromBinary was unchanged (and it looks like the size of the binary needs to be serialized to be able to deserialize)

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/ingest/GeometrySimpOptionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+			Geometry geom ) {
+		return (geom.getCoordinates().length < this.maxVertices && !geom.isEmpty() && geom.isValid());
+	}
+
+	@Override
+	public byte[] toBinary() {
+		return new byte[] {
+			(byte) maxVertices,
+			(byte) simpVertMin,
+			(byte) tolerance
```
rfecher: casting to a byte isn't going to work

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/ingest/GeometrySimpOptionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+		};
+	}
+
+	@Override
+	public void fromBinary(
+			byte[] bytes ) {
+		int offset = 0;
+		maxVertices = (int) bytes[offset];
```
rfecher: and the array would be of size 3 if it did...suggest to use a bytebuffer

extensions/adapters/vector/src/main/java/mil/nga/giat/geowave/adapter/vector/ingest/SimpleFeatureIngestOptions.java
```diff
@@ -30,11 +30,23 @@
...
 
+	@ParametersDelegate
+	private GeometrySimpOptionProvider simpOptionProvider = new GeometrySimpOptionProvider();
+
 	@ParametersDelegate
 	private Object pluginOptions = null;
 
 	public SimpleFeatureIngestOptions() {}
 
+	public GeometrySimpOptionProvider getSimpOptionProvider() {
```
rfecher: a nit, but can "geom" or "geometry" be somewhere in the method names?  it seems like an important descriptor

extensions/formats/gpx/src/main/java/mil/nga/giat/geowave/format/gpx/GpxIngestPlugin.java
```diff
@@ -313,4 +317,9 @@ public IngestGpxTrackFromHdfs(
 			Time.class
 		};
 	}
+
+	public void setExtentOptionProvider(
```
rfecher: is this called anywhere? Just trying to figure out where/how the JCommander annotations for this option provider get wired in.

JWileczek: This is called from the GpxIngestFormat. I'm not really a fan of how I had to create the OptionProvider within the plugin + format. Ultimately, the JCommander annotations get set via SimpleFeatureIngestOptions because the Gpx specifc  plugin options get added via the internalGetIngestFormatOptionProviders. SimpleFeatureIngestOptions has the parameter delegate annotation set here. I attempted to attach it via GpxIngestPlugin originally, but since the plugin is created/returned as a AbstractSimpleFeatureIngestPlugin it wasn't picking up that annotation in the GpxIngestPlugin.
@ParametersDelegate
private Object pluginOptions = null;

PR #1350, opened 2018-06-21
REST services parameter handling enhancements

dbazile:
An attempt to smooth out some of the edges around the REST API and how it handles parameter parsing, specifically:

- Fixes errors from unconverted assignments of primitive `Object[]` arrays to fields
- Improves error logging for diagnostics/troubleshooting
- Adds null checks for optional/missing parameters


dbazile:
Arg, mis-clicked. :)

PR #1358, opened 2018-07-19
reorder REST services parameter converter

cjw5db:


services/rest/src/main/java/mil/nga/giat/geowave/service/rest/GeoWaveOperationServiceWrapper.java
```diff
@@ -189,63 +189,71 @@ private void injectParameters(
...
-					else if (Float.class.isAssignableFrom(type) || float.class.isAssignableFrom(type)) {
-						objValue = Float.valueOf(strValue);
-					}
-					else if (Boolean.class.isAssignableFrom(type) || boolean.class.isAssignableFrom(type)) {
-						objValue = Boolean.valueOf(strValue);
+			Field field = f.getField();
+			final String strValue = (String) requestParameters.getString(f.getName());
+			
+			if (field.isAnnotationPresent(Parameter.class)) {
+				Class<? extends IStringConverter<?>> converter = field.getAnnotation(
```
emacthecav: So right here, the code gets the Class object and tries to see if it is one of the converters, or if it's "NoConverter." It does this for every single field, even though only a few have converters, which seems inefficient to me. However, it is important to do this before some of the other checks, because there are certain cases where the other checks would catch the type. I.E. passwordCoverter converts a string to another string, so that would get caught in the String check

rfecher: seems reasonable enough, I'd check for converter == null at the front of the if statement even though apparently it must always get set to "NoConverter" in practice just to add a little more protection from NPEs

services/rest/src/main/java/mil/nga/giat/geowave/service/rest/field/StringMainParam.java
```diff
@@ -27,4 +27,7 @@ protected String valueToString(
 		return value;
 	}
 
+	public Field getField() {
+		return super.getField();
```
rfecher: why bother to override the method if its just going to delegate with `super`?

services/rest/src/main/java/mil/nga/giat/geowave/service/rest/field/ListMainParam.java
```diff
@@ -30,5 +30,9 @@ protected String valueToString(
 				value,
 				',');
 	}
+	
+	public Field getField() {
```
rfecher: why bother to override the method if its just going to delegate with `super`?

PR #1376, opened 2018-08-20
Small update to bootstrap scripts to fix original jupyter bootstrap

JWileczek:
Small fix to original jupyter bootstrap script after jupyterhub changes.


JWileczek:
Closing in favor of wrapping these changes into UDT PR. Minor enough that they shouldn't require their own PR

PR #1421, opened 2018-09-17
Updating spatial join test and pyspark module settings

JWileczek:
Spatial join test is now correctly pulling from two stores instead of one.
Pyspark module is always included in project/build, but python package only built if python profile is attached.

### DIFF
--- a/analytics/pyspark/pom.xml
+++ b/analytics/pyspark/pom.xml
@@ -42,10 +42,18 @@
                     </execution>
                 </executions>
             </plugin>
+            </plugins>
+        </build>
 
+    <profiles>
+    <profile>
+    <id>python</id>
+    <build>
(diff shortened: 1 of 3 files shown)


analytics/pyspark/pom.xml
```diff
@@ -42,10 +42,18 @@
                     </execution>
                 </executions>
             </plugin>
+            </plugins>
+        </build>
 
+    <profiles>
+    <profile>
+    <id>python</id>
```
rfecher: @JWileczek I have a PR coming, but I'm pretty sure the profile has to be "pyspark" or we need to change the profile being activated from deploy, right?  Well, thats what I assume and am going to just name this profile "pyspark" so everything should work the same.

PR #1465, opened 2018-11-19
fix for optimalcqlquery within distributed processing

rfecher:


rfecher:
this is replaced by PR #1473

Issue #1466, opened 2018-11-28
Hbase bulk delete fails with InsertionIdQueryFilter


richard3d:
When the bulk delete request is made to the endpoint it seems that the MultiRowRangeFilter and the InsertionIdQueryFilter are incompatible. It seems a work around is to check for single row filters in the 'HBaseOperations' class's 'createDeleter' method.

rfecher:
Fixed

PR #1467, opened 2018-11-28
Duplicate deletion IT and implementation

richard3d:
Added test to SpatialTemporalQueryIT for checking deletion of duplicate entries across time ranges, and implemented duplicate deletion by insertion ID via deletion callback. This PR also addresses bulk deletion incompatibility with InsertionIdQueryFilter referenced by issue #1466.


core/store/src/main/java/org/locationtech/geowave/core/store/api/DataStore.java
```diff
@@ -105,6 +107,27 @@
...
+	 * @param query
+	 *            data constraints for the query and additional options for
+	 *            processing the query
+	 * @param filterDuplicates
+	 *            should the iterator filter out duplicates?
+	 * @return An iterator on all results that match the query. The iterator
+	 *         implements Closeable and it is best practice to close the
+	 *         iterator after it is no longer needed.
+	 */
+	<T> CloseableIterator<T> query(
```
rfecher: I am _really_ cautious about adding anything to `DataStore` because we are really trying hard to keep this interface as clean as possible.  In this case this method on the `DataStore` interface seems pretty unnecessary unless I'm missing something.  The `DuplicateDeletionCallback` can use `BaseDataStore` and it only needs access to `delete(Query, deleteDuplicates)` which in turn merely needs an extra flag on `internalQuery` and not the additional `query` method.  The only place I see the query used with the boolean for filterDuplicates is within the ITs which fundamentally I try not to make major code impacts to satisfy something in the ITs, to the extent possible.  In this case, it seems to me there's a couple of options available to the ITs to do the exact same thing - one is to just cast to BaseDataStore in the IT and assume these are all BaseDataStores that are being tested. Thats easiest but a bit more intrusive than what may be a little more work but completely unintrusive - within the IT you could also just overwrite the `DuplicateEntryCount` within the DataStatisticsStore to be 0 when you don't want to dedupe on query and 1 or whatever it was originally when you want it to behave normally on query/delete.  Its a hack, but it would be a hack in the IT that should get the job done without needing any extra code in the baseline.

test/src/test/java/org/locationtech/geowave/test/query/SpatialTemporalQueryIT.java
```diff
@@ -446,6 +462,32 @@ private static void ingestTimeRangeData(
...
+				min);
+		featureTimeRangeBuilder.add(geomFactory.createPoint(new Coordinate(
+				0,
+				0)));
+		featureTimeRangeBuilder.add(cal.getTime());
+		cal.set(
+				field,
+				max);
+		featureTimeRangeBuilder.add(cal.getTime());
+		SimpleFeature feature = featureTimeRangeBuilder.buildFeature(name + ":fullrange");
```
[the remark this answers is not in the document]
richard3d: Changed the name to avoid headaches down the road. This was one of those concepts that was a bit fuzzy to me originally (which data needs to be unique), technically it still worked. But, queries with dataId would fail in this case.

core/store/src/main/java/org/locationtech/geowave/core/store/base/BaseDataStore.java
```diff
@@ -232,13 +249,27 @@ public DataStatisticsStore getStatisticsStore() {
...
+		// Note: filterDuplicates has a different effect depending if delete is
+		// true or not.
+		// If delete is false and filterDuplicates is true, the query will
+		// return
+		// just the entries from the tables corresponding to the specified range
+		// If delete is false and filterDuplicates is false, the query will
+		// return all the entries (duplicates included)
+		// if delete is true and filterDuplicates is true, the query will filter
+		// duplicates, but the duplicate deletion callback WILL NOT BE added
+		// if delete is true and filterDuplicates is false, the query will
```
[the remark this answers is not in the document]
richard3d: I can tighten up the language here but the underlying problem is this: If you perform a deletion query on a partial range (e.g. 1970-1971, full range is 1970-1975, dupes are 1972-1975), it will only delete the partial range (1970-1971) regardless if that filter is null or not. This is what I meant by "duplicates still get filtered", I suppose I should say "they won't get deleted". According to the logic in `BaseConstraintsQuery` the dedupe filter is only applied client side as well, hence why it has no effect when delete is called and why I created the duplicate deletion scan callback. I updated the comment, but let me know if you think there is a better path or way to implement this.

core/store/src/main/java/org/locationtech/geowave/core/store/api/DataStore.java
richard3d: Totally agree, I added this originally for IT but thought it might be handy for users if they ever needed a sanity check on their data as well. I'll use a work around for the IT instead and remove.

core/store/src/main/java/org/locationtech/geowave/core/store/callback/DuplicateCountCallback.java
```diff
@@ -0,0 +1,58 @@
...
+import org.locationtech.geowave.core.store.api.QueryBuilder;
+import org.locationtech.geowave.core.store.base.BaseDataStore;
+import org.locationtech.geowave.core.store.callback.DeleteCallback;
+import org.locationtech.geowave.core.store.entities.GeoWaveRow;
+import org.locationtech.geowave.core.store.query.constraints.InsertionIdQuery;
+import org.locationtech.geowave.core.store.util.DataStoreUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
```
rfecher: This file is only used in the ITs? In which case would it make more sense to have it in the ITs even as a static private inner  class? Also, minor nit is the javadoc header looks to be a copy of the DuplicateDeletionCallback and not accurate for this class.

PR #1501, opened 2019-02-09
CopyTo test and deleteall fix

buman56:
Small test for getTypes, getIndices, and copyTo

Line added to make dataStore.deleteAll() to work properly


rfecher:
@buman56 there is a conflict with master (from code from your teammates).  can you rebase with master and resolve the conflicts?

PR #1502, opened 2019-02-09
Query test for DataStore

laiterry:
Unit Test for query method of DataStore.
QueryTestFailed indicates that there may be a bug in adding same index to different adapters . dataStore.addType(adapter, index);
dataStore.addType(adapter2, index);


rfecher:
The fix for multiple adapters in the same index is contained in `NativeEntryIteratorWrapper`.  Please replace with this and give it a shot: https://gist.github.com/rfecher/cdfd0d1e9304e75fa1787bb0d4a44988

rfecher:
@laiterry see the comment above and the gist: https://gist.github.com/rfecher/cdfd0d1e9304e75fa1787bb0d4a44988

PR #1503, opened 2019-02-09
Use xenial (ubuntu 16.04) for Travis environment

dannyqiu:
`dist: precise` is deprecated


nikmanG:
might be prudent to also add `dist: trusty` into the matrix since it's only being deprecated in April.

PR #1505, opened 2019-02-10
Add tests for hbase util

foolhb:
Add the most basic unit test for hbase util class.
Have some difficulty in mocking hbase connection, so most of the classes are not tested


foolhb:
Wrong mail and username

PR #1506, opened 2019-02-10
Add tests for hbase util

foolhb:
Add basic unit tests for hbase extension.


foolhb:
Wrong eclipse agreement

PR #1507, opened 2019-02-10
Add basic unit tests for hbase extension.

foolhb:
Add basic unit tests for hbase extension.


foolhb:
ip validation fails

PR #1509, opened 2019-02-10
Add hbase util test

foolhb:


foolhb:
extensions/datastores/hbase/pom.xml

PR #1510, opened 2019-02-10
Add hbase extension tests

foolhb:


extensions/datastores/hbase/pom.xml
```diff
@@ -1,39 +1,70 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
rfecher: I suggest we undo these pom changes.  It looks like the tests have no need for minihbase cluster which is good.  Can we just commit the tests without this pom change?

PR #1511, opened 2019-02-10
Stats unit test

alexanderGGreenberg:
Most basic unit unit tests.
Signed-off-by: Alexander Greenberg <agg73@cornell.edu>

### DIFF
--- a/extensions/adapters/vector/src/test/java/org/locationtech/geowave/adapter/vector/stats/VectorStatsTest.java
+++ b/extensions/adapters/vector/src/test/java/org/locationtech/geowave/adapter/vector/stats/VectorStatsTest.java
@@ -1,0 +1,90 @@
+package org.locationtech.geowave.adapter.vector.stats;
+
+import java.util.ArrayList;
+import java.util.Map;
+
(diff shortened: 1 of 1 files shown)


extensions/adapters/vector/src/test/java/org/locationtech/geowave/adapter/vector/stats/VectorStatsTest.java
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	
+	@Test
+	public void test(){
+		StatisticsQuery<Envelope> query= 
+		VectorStatisticsQueryBuilder.newBuilder().factory().bbox().build();
+		Statistics<Envelope>[] stat = dataStore.queryStatistics(query);
+		Assert.assertEquals(stat[0].getResult().getMaxX(), 5000,0);
+		Assert.assertEquals(stat[1].getResult().getMaxX(), 55,0);
```
rfecher: order of the stat results is not guaranteed.  You should check the getDataTypeName() from the statistics returned to match the appropriate statistic with the expected value for each feature type.

PR #1516, opened 2019-02-10
Python bindings

nikmanG:
Pom reqs added and minimum exposed API.


rfecher:
this has conflicts that need to be resolved

PR #1517, opened 2019-02-10
add kudu example

carolyntang:


extensions/datastores/kudu/pom.xml
```diff
@@ -12,6 +12,51 @@
...
 	<description>Geowave Data Store on Apache Kudu</description>
 
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>2.3.1</version>
+				<configuration>
+					<source>1.7</source>
```
rfecher: we're going to want to be on 1.8

PR #1518, opened 2019-02-11
Create removal tests and necessary fixes.

nikmanG:
The equals and hashCode in `GeoWaveMetadataWithUUID` were created because of an error that happened from spotbugs. Everything else was a result from tests.


nikmanG:
Not sure why ip-validation is throwing an error because both committer and author are signed off as being me https://api.github.com/repos/nikmang/geowave/commits/a54821b3b7481c603d028b0f54d33ce6d2869431

core/store/src/test/java/org/locationtech/geowave/core/store/api/DataStoreTestSuite.java
```diff
@@ -0,0 +1,17 @@
...
+ * under the terms of the Apache License, Version 2.0 which accompanies this distribution and is
+ * available at http://www.apache.org/licenses/LICENSE-2.0.txt
+ */
+package org.locationtech.geowave.core.store.api;
+
+import org.junit.runner.RunWith;
+import org.junit.runners.Suite;
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({DataStoreAddTest.class, DataStoreRemoveTest.class})
```
rfecher: can you delete this?  This is unnecessary - I think the maven plugin we setup to actually only run unit tests named *Test which is actually good because otherwise these tests would run twice (once themselves and once when it encounters this suite).  Anyways, better off not to worry about this.

core/store/src/test/java/org/locationtech/geowave/core/store/api/DataStoreRemoveTest.java
```diff
@@ -0,0 +1,222 @@
...
+public class DataStoreRemoveTest {
+  private static int counter = 0;
+  private static final String MOCK_DATA_TYPE_1 = "Some Data Type";
+  private static final String MOCK_DATA_TYPE_2 = "Another Data Type";
+
+  private DataStore dataStore;
+
+  @Before
+  public void createStore() {
+    dataStore = DataStoreFactory.createDataStore(new MemoryRequiredNameSpaceOptions(counter + ""));
```
rfecher: now that it is properly deleting everything, can we clean this up and remove this and associated inner class

nikmanG: Fixed in commit.

core/store/src/test/java/org/locationtech/geowave/core/store/api/DataStoreTestSuite.java
nikmanG: Fixed in commit.

PR #1530, opened 2019-03-14
Fix of data missing for testing Sentinel2 extension

ahuarte47:
This PR adds a missing JP2 file used for testing in AWS Sentinel2 provider of Sentinel cli extension


ahuarte47:
Thanks @rfecher !

PR #1541, opened 2019-04-21
SimpleFeature support & python packaging

charlesyu108:


sinistersnare:
Python packages dont start with src/main, so it can be changed ` python/src/main/python/pygw/` to `python/pygw`. Or even just leave it at `pygw`.

charlesyu108:
This PR moves @alexanderGGreenberg's work into geotools.py and adds support for the creation of  SimpleFeatureAdapter and SimpleFeature instances.

PR #1558, opened 2019-05-10
Solved a bug in ZOrderSFC.getRange and add ZOrderSFCTest

LB-Yu:
Hi!
I found that the *getRanges* function of class *ZOrderSFC* has a bug. In the original code, the return value of the function is always constant, which is the maximum and minimum of each dimension. For example, when I encode latitude and longitude, its return value is always [[-180.0, 180.0], [-90.0, 90.0]].

This bug is caused by the *decodeRanges* function and the *decode* function of the class *ZOrderUtils*. I have solved this bug and added test code for the class *ZOrderSFC*.


rfecher:
there's an issue with the maven repo that geotools is hosted on, we have to wait for [this](https://trac.osgeo.org/osgeo/ticket/2261) to be resolved.  In the meantime can you please sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php).  The ECA is required for all contributions.  Thanks!

LB-Yu:
I've done that now!

PR #1566, opened 2019-05-24
Implement KDE and Raster Resize in Spark

rfecher:


analytics/api/src/main/java/org/locationtech/geowave/analytic/kryo/GridCoverageWritableSerializer.java
```diff
@@ -0,0 +1,49 @@
+package org.locationtech.geowave.analytic.kryo;
```
jdgarrett: Missing header.

analytics/spark/src/main/java/org/locationtech/geowave/analytic/spark/GeoWaveRDDLoader.java
```diff
@@ -134,4 +136,59 @@ public static GeoWaveIndexedRDD loadIndexedRDD(
...
+
+    if (indexName != null) {
+      GeoWaveInputFormat.setQuery(
+          conf,
+          QueryBuilder.newBuilder().indexName(indexName).build(),
+          storeOptions.createAdapterStore(),
+          storeOptions.createInternalAdapterStore(),
+          storeOptions.createIndexStore());
+    }
+    if (((minSplits != null) && (minSplits > -1)) || ((maxSplits != null) && (maxSplits > -1))) {
```
jdgarrett: Should this be `&&` instead of `||`, or is it fine to specify only `minSplits` or `maxSplits`?

analytics/spark/src/main/java/org/locationtech/geowave/analytic/spark/GeoWaveRasterRDD.java
```diff
@@ -0,0 +1,33 @@
+package org.locationtech.geowave.analytic.spark;
```
jdgarrett: Missing header.

analytics/spark/src/main/java/org/locationtech/geowave/analytic/spark/RDDUtils.java
```diff
@@ -189,43 +199,86 @@ public static InsertionIds trimIndexIds(
...
   }
+
+  public static void writeRasterRDDToGeoWave(
+      final SparkContext sc,
+      final Index index,
+      final DataStorePluginOptions outputStoreOptions,
+      final RasterDataAdapter adapter,
+      final JavaRDD<GridCoverage> inputRDD) throws IOException {
+
+    writeRasterToGeoWave(sc, index, outputStoreOptions, adapter, inputRDD);
```
jdgarrett: Is there a need for the private method to exist?  It has the same signature as the public one.

analytics/spark/src/main/java/org/locationtech/geowave/analytic/spark/kde/KDERunner.java
```diff
@@ -0,0 +1,734 @@
+package org.locationtech.geowave.analytic.spark.kde;
```
jdgarrett: Missing header.

analytics/spark/src/main/java/org/locationtech/geowave/analytic/spark/kde/operations/KDESparkOptions.java
```diff
@@ -0,0 +1,152 @@
+package org.locationtech.geowave.analytic.spark.kde.operations;
```
jdgarrett: Missing header.

analytics/spark/src/main/java/org/locationtech/geowave/analytic/spark/resize/RasterTileResizeSparkRunner.java
```diff
@@ -0,0 +1,248 @@
+package org.locationtech.geowave.analytic.spark.resize;
```
jdgarrett: Missing header.

core/store/src/main/java/org/locationtech/geowave/core/store/cli/remote/RemoteCopyCommand.java
```diff
@@ -0,0 +1,67 @@
+package org.locationtech.geowave.core.store.cli.remote;
```
jdgarrett: Missing header.

core/store/src/main/java/org/locationtech/geowave/core/store/entities/GeoWaveRowMergingTransform.java
```diff
@@ -0,0 +1,37 @@
+package org.locationtech.geowave.core.store.entities;
```
jdgarrett: Missing header.

core/store/src/main/java/org/locationtech/geowave/core/store/util/DataStoreUtils.java
```diff
@@ -413,13 +419,18 @@ public static ByteArray ensureUniqueId(final byte[] id, final boolean hasMetadat
...
+                + vis2.length);
+    buffer.put(OPEN_PAREN_BYTE);
     buffer.put(vis1);
-    buffer.putChar(')');
-    buffer.put(BEG_AND_BYTE);
+    buffer.put(MERGE_VIS_BYTES);
     buffer.put(vis2);
-    buffer.put(END_AND_BYTE);
+    buffer.put(CLOSE_PAREN_BYTE);
     return buffer.array();
```
jdgarrett: Are we worried about something like `vis1` being `(a)&(b)` as a result of a previous merge and `vis2` being `a`?  After a few calls of passing `a` as `vis2` with the result of the previous call as `vis1`, you'd end up with something like `((((a)&(b))&(a))&(a))&(a)`.  Perhaps it would be best to implement the merging logic into the `VisibilityExpression` class and only merge when `vis2` is not already satisfied by `vis1`.

analytics/spark/src/main/java/org/locationtech/geowave/analytic/spark/RDDUtils.java
rfecher: good catch, I think this went through several revisions and with what was left the extraneous method makes no sense

core/store/src/main/java/org/locationtech/geowave/core/store/util/DataStoreUtils.java
rfecher: thats a fair point - I actually just fixed the serialization from what was blatantly a buffer overflow in the previous implementation under-allocating the bytebuffer, but there's definitely room for improvement here.  I think your suggestion is good, I'll fix it.

rfecher: after thinking a little more, there is a complication ... your suggestion would work if you can find one of the visibilities to be a single token so you can basically treat it as an authorization to evaluate, but doesn't extend to the case where both visibilities are boolean logic expressions

rfecher: I _think_ the correct way of handling this is to *always* normalize all visibility expressions all the time following some consistent and repeatable criteria.  Accumulo I think has a pretty good way of normalizing expressions: https://github.com/apache/accumulo/blob/2b1c4d008a7e868f0cb54f79c7b1613685b98921/core/src/main/java/org/apache/accumulo/core/security/ColumnVisibility.java#L217-L247

I'd say we normalize expressions as they come in and before they are stored, and normalize expressions after they are merged.  That would seem to solve the problem here in both the easy case where you can find one visibility thats a single token as well as the more complex generic case.

rfecher: however, I'll happily say that I think this one is outside of the scope of this particular PR and suggest it as a separate issue.  I created #1574

PR #1569, opened 2019-05-29
fix for simple query deletion on secondary index

rfecher:


rfecher:
this fix is applied to the data store index auto-selection PR #1572

PR #1572, opened 2019-05-31
datastore query auto-select best index

rfecher:


core/store/src/main/java/org/locationtech/geowave/core/store/base/BaseQueryOptions.java
```diff
@@ -430,14 +430,37 @@ public void setAggregation(
    * @return
```
jdgarrett: Javadoc should be updated.

core/store/src/main/java/org/locationtech/geowave/core/store/callback/DeleteOtherIndicesCallback.java
```diff
@@ -0,0 +1,83 @@
+package org.locationtech.geowave.core.store.callback;
```
jdgarrett: Missing header.

test/src/test/java/org/locationtech/geowave/test/secondary/SimpleQuerySecondaryIndexIT.java
```diff
@@ -0,0 +1,151 @@
+package org.locationtech.geowave.test.secondary;
```
jdgarrett: Missing header.

PR #1595, opened 2019-06-28
Feature/RAD-1219

aibrahimRS:
Removed any reference of Sandbox VM and added in Stand-Alone Installer Updates


docs/content/standalone-installer/001-hw-quickstart-guide-setup.adoc
```diff
@@ -0,0 +1,11 @@
+
+<<<
+
+_v${project.version}_
+
+= GeoWave Quickstart Guide Stand-Alone Installer Setup image:geowave-icon-logo-cropped.png[width="28%"]
+
+== Introduction
+
+The Stand-Alone Installer is meant to be a replacement for the Sandbox environemnt. The GeoWave Installer comes with many services such as Accumulo, Hbase, and Cassandra. These services are used as database management systems in order to handle large amounts of data. The installer doesnt come with services such as Apache Maven and Hadoop which are required for some of the commands needed in the Demos. These services can be installed and setup from the links below. The installer is continously being updated on a constant basis with the links provided in the User Guide.
```
rfecher: fix typos in here
also, maven is not required at all, and the hadoop binaries are only needed for windows. We should be able to add them to the installer so it doesn't require an extra step.

rfecher:
it looks like instructions were changed for EMR and the sandbox specific instructions were removed.  We can't change any instructions for EMR otherwise that guide won't work.

Also, we should be able to delete any of the images in the "images/sandbox" directory.

PR #1607, opened 2019-08-13
Fixed logic error in close method of RocksDBClient

blastarr:


blastarr:
I'll let Rich do this since I'm out of date.

PR #1609, opened 2019-08-14
Upgraded HBase2 API

combineads:
Signed-off-by: Byunghwa Yun combine@combineads.co.kr

### DIFF
--- a/deploy/pom.xml
+++ b/deploy/pom.xml
@@ -588,6 +588,32 @@
 										<excludes>
 											<exclude>org.slf4j:*</exclude>
 											<exclude>com.google.protobuf:*</exclude>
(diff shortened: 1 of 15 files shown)


rfecher:
thanks! it looks like a number of tests have failed (although a few of travis-ci's jobs did pass fully) and in general with our 1.0.0 release we need to maintain backwards compatibility with hbase 1.x (particularly versions >= 1.2  and < 2.0).  This will be a great resource for us to move to hbase 2 after the geowave 1.0.0 release, but we will need tests passing (and if we make the move prior to geowave 1.0.0 release, we will have to find a way for backwards compatibility which may simply not be possible).

rfecher:
PR #1813 covered this

PR #1617, opened 2019-09-04
Brute Force Merging for HBase and CLI partition strategy converter

rfecher:


core/store/src/main/java/org/locationtech/geowave/core/store/operations/remote/options/BasicIndexOptions.java
```diff
@@ -38,4 +42,22 @@ public PartitionStrategy getPartitionStrategy() {
   public void setPartitionStrategy(final PartitionStrategy partitionStrategy) {
     this.partitionStrategy = partitionStrategy;
   }
+
+  public static class PartitionStrategyConverter implements IStringConverter<PartitionStrategy> {
+
+    @Override
+    public PartitionStrategy convert(final String value) {
+      final PartitionStrategy convertedValue = PartitionStrategy.fromString(value);
```
jdgarrett: Is there a reason `PartitionStrategy.valueOf` couldn't be used instead of making a new function?

PR #1631, opened 2019-10-10
delete pyramid levels of raster data

rfecher:


extensions/adapters/raster/src/main/java/org/locationtech/geowave/adapter/raster/operations/DeletePyramidLevelCommand.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+    inputStoreOptions = inputStoreLoader.getDataStorePlugin();
+    final DataStore store = inputStoreLoader.createDataStore();
+    RasterDataAdapter adapter = null;
+
+    for (final DataTypeAdapter<?> type : store.getTypes()) {
+      if (isRaster(type)
+          && ((coverageName == null) || coverageName.equals(adapter.getTypeName()))) {
+        if (adapter != null) {
+          LOGGER.error(
```
jdgarrett: There is some inconsistency with the way status/errors are being displayed to the user.  In other CLI commands errors are typically reported via ParameterException, with information type text being reported via `JCommander.getConsole().println`.  I think the error logging should be updated and perhaps the `LOGGER.info` parts as well, but as I said before, there seems to be some inconsistency amongst existing commands.

extensions/adapters/raster/src/main/java/org/locationtech/geowave/adapter/raster/operations/DeletePyramidLevelCommand.java
```diff
@@ -0,0 +1,222 @@
...
+      if (!pStats.getPartitionKeys().remove(p)) {
+        LOGGER.error(
+            "Unable to remove partition " + p.getHexString() + " for pyramid level " + level);
+        return;
+      }
+    }
+    statsStore.setStatistics(ovStats);
+    statsStore.setStatistics(pStats);
+
+    for (final ByteArray p : partitions) {
```
jdgarrett: Is it possible that there could be no statistics for the given coverage, but still have data that could be removed?

extensions/adapters/raster/src/main/java/org/locationtech/geowave/adapter/raster/operations/DeletePyramidLevelCommand.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return adapter instanceof RasterDataAdapter;
+  }
+
+  public List<String> getParameters() {
+    return parameters;
+  }
+
+  public void setParameters(final String inputStore, final String outputStore) {
+    parameters = new ArrayList<>();
```
jdgarrett: This command only expects a single parameter.

extensions/adapters/raster/src/main/java/org/locationtech/geowave/adapter/raster/operations/DeletePyramidLevelCommand.java
rfecher: I looked at this and I think there's a couple ways to look at it - always throw an exception on errors would be an option and let GeoWaveMain take care of handling exceptional cases, or always use JCommander.getConsole().println() although because a lot of logic in geowave commands is potentially shared with REST and gRPC services this is probably not the right approach.  It seems actually we should be moving more to some generalized logging that just happens to be to JCommander console from the CLI but is logging to file for the services.  Regardless, I think whatever is done, it may as well be done consistently outside of the scope of this.

extensions/adapters/raster/src/main/java/org/locationtech/geowave/adapter/raster/operations/DeletePyramidLevelCommand.java
rfecher: good catch

extensions/adapters/raster/src/main/java/org/locationtech/geowave/adapter/raster/operations/DeletePyramidLevelCommand.java
rfecher: yeah, I think the only somewhat legit case may be is statistics are disabled by the options, and I added a check for that (although with raster data functionality would be greatly diminished if stats are disabled so its probably something that just shouldn't happen, the overview stat is used to pick a resolution if you want to composite an image based on geobounds and pixel height/width, without that you would primarily be needing to query for tiles and composite them into an image externally).

Otherwise if there stats aren't found, something is wrong and it stands to reason you don't want to delete the data without properly maintaining the stats and get in a more chaotic situation.  The only thing I can think of is to add a `--force` flag that does it anyways, but probably outside the scope here.

PR #1633, opened 2019-10-10
limited implementation of pygw stats

rfecher:
at least it gets count() working reasonably well

### DIFF
--- a/python/src/main/python/pygw/query/statistics/noop_transformer.py
+++ b/python/src/main/python/pygw/query/statistics/noop_transformer.py
@@ -1,0 +1,14 @@
+from ..query_result_transformer import QueryResultTransformer
+#from py4j.java_gateway import JavaGateway
+
+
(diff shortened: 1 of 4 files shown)


python/src/main/python/pygw/query/statistics/noop_transformer.py
```diff
@@ -0,0 +1,14 @@
+from ..query_result_transformer import QueryResultTransformer
+#from py4j.java_gateway import JavaGateway
+
+
+class NoOpTransformer(QueryResultTransformer):
+    """
+    Transforms Java Long into a pyhton
```
jdgarrett: This comment seems to be inaccurate.  It actually transforms any primitive java type to the python variant, and if a python variant couldn't be found, it will return the java reference.

jdgarrett:
Would it be difficult to add some tests to cover this additional functionality?  Pygw has very high test coverage and I would like to stay on top of it.

PR #1641, opened 2019-12-11
Add some docs for core classes

Bluefire2:
Add javadocs for {Metadata, Row} x {Reader, Writer, Deleter}. We also removed `public` from interface methods since it's redundant (interface methods can't have any other access level in Java 8).


Bluefire2:
How could the coverage decrease (or change at all), if only docstrings are changed? Was there a concurrent PR that had been merged?

jdgarrett:
Looks like the formatter needs to be run on the updated classes before the build will pass.

jdgarrett:
Changed files can be formatted by running `mvn clean install` from the GeoWave directory.  Once the files in question have been run through the formatter, the build will pass and this PR can be merged.

jdgarrett:
I have rolled the changes from this PR into #1669.

PR #1650, opened 2019-12-30
Add new store commands; genericize console output

jhickman-prominent:
Per discussion with JD.


core/cli/src/main/java/org/locationtech/geowave/core/cli/utils/FirstElementListComparator.java
```diff
@@ -0,0 +1,41 @@
+package org.locationtech.geowave.core.cli.utils;
```
jdgarrett: Missing copyright header

jhickman-prominent: Added missing copyright header

Issue #1652, opened 2019-12-31
Provide Easier Mechanics for Custom Indexing


rfecher:
One may want to be able to just index their data given some custom logic (such as using a set of enumerated values or a string, or really anything that isn't provided by the numeric index strategy as it is now).  While there are ways currently to add this logic, the third-party code is not as straightforward as it could be.  Ideally there is can just be an interface that needs to be implemented (with arbitrary provided logic) to provide `InsertionIds` from entries and to provide `QueryRanges` from a known constraint type, then queries should be able to work on that index that match the known constraint type.

jdgarrett:
Resolved with #1653

Issue #1651, opened 2019-12-31
Inconsistency when Indexing Duplicates in Redis 


rfecher:
When an index produces duplicates for a single result, Redis only allows one entry in a collection with the same value (so there may be a different key, but the purpose of duplicates is to have the exact same value).  Redis behavior is to overwrite with the most recent matching value within the collection.  To maintain all backwards compatibility but allow for duplicates to behave properly, there is an optional, nillable `duplicateId` that is persisted in the collection (assuming that the redis writer is given an array of geowave rows with all duplicates, if this assumption is wrong, at best we could just generate a random duplicate ID if the other duplicate values are not in context at the same time).

PR #1653, opened 2019-12-31
added custom index and fixed issue with redis when there are duplicates (issues #1651 and #1652)

rfecher:
this addresses #1651 and #1652

`CustomIndexStrategy` is the simple interface that can be implemented to insert your own index strategy, with an example and test in GeoWaveCustomIndexIT for indexing arbitrary enums


test/src/test/java/org/locationtech/geowave/test/basic/GeoWaveCustomIndexIT.java
```diff
@@ -0,0 +1,340 @@
+package org.locationtech.geowave.test.basic;
```
jdgarrett: Missing header

Issue #1655, opened 2020-01-02
Multi-Index Deletion Maintaining IndexMetadata


rfecher:
There's an issue with deletion where the IndexMetadataSet delete callback assumes the GeoWaveRows are from that index when they can be coming from a different index.  So line 111 of IndexMetaDataSet can be inaccurately passing an indexmetadata insertion IDs that don't belong to the associated index.  This can result in incorrect updates at line 609 of TieredSFCIndexStrategy of the metadata, to include potentially index out of bounds exception (if the incoming insertion IDs doesn't have a partition ID, even though the tiered index metadata assumes a partition ID because any insertion ID from that index will have the tier ID).  See commented out test code within GeoWaveCustomIndexIT which should expose this issue.

jdgarrett:
Fixed with #1670

PR #1664, opened 2020-01-10
fixed issue with deleting from custom index

rfecher:
There was a relatively minor issue simply when deleting from a custom index that was originally masked by issue #1655 and the multi-index deletion issues.  Added a test for the custom index query and deletion all by itself along with another test for deletion with the custom index, spatial, and spatial-temporal indices.  The commented line with the TODO still exists with the multi-index deletion bug, but this resolves simply deletion from the custom index.


jdgarrett:
It looks good to me, but there are test failures in many of the IT profiles of the build matrix.

PR #1715, opened 2020-05-01
Fix for #1702, additional tests for GeoWaveFeatureReader, doc updates.

### LINKED ISSUE #1702: [Bug] Time Bounds Handling through GeoWaveFeatureReader
Logic in the [QueryIndexHelper](https://github.com/locationtech/geowave/blob/689387635e0ca129efeb41c727896d315507ad1b/extensions/adapters/vector/src/main/java/org/locationtech/geowave/adapter/vector/util/QueryIndexHelper.java) produces undesired results through GeoServer. If the bounds of the query fall outside the bounds of available data for the feature, based on the TimeRangeStatistics, it removes the temporal constraints from the query, instead of producing no results. This is documented in the TODO of the clipIndexedTemporalConstraints function.

richard3d:
Fixes #1702 and also handles cases where the geometry constraint is empty. Fixes a previously unknown bug with queries based on a temporal constraint only that broke depending on if the user issued the query with ECQL vs CQL. Added test for queries that only use temporal constraint to catch this issue in the future should it arise again. Also includes dev-guide update for debug configurations, fixing errors from pom.xml for plugin execution.


richard3d:
Forgot to format and signoff will re-issue

PR #1734, opened 2020-06-05
added default to/from binary to aggregation

rfecher:
I know others are in this code so this may be more of a merge head-ache than its worth (take it or leave it), just a simple suggestion to clarify the to/from binary of aggregation based on @jdgarrett's recent comment.


sinistersnare:
Yeah, this looks good to me too. After this gets in, I'll rebase my MR on top of it so I can remove the same code.

PR #1788, opened 2021-01-21
GEOWAVE-1787: enumerated field indexing support

rfecher:
Signed-off-by: Rich Fecher <richard.fecher@maxar.com>

### DIFF
--- a/test/src/test/java/org/locationtech/geowave/test/basic/GeoWaveEnumIndexIT.java
+++ b/test/src/test/java/org/locationtech/geowave/test/basic/GeoWaveEnumIndexIT.java
@@ -13,6 +13,7 @@ import java.io.InputStreamReader;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVRecord;
 import org.geotools.feature.simple.SimpleFeatureBuilder;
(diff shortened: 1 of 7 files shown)


test/src/test/java/org/locationtech/geowave/test/basic/GeoWaveEnumIndexIT.java
```diff
@@ -0,0 +1,231 @@
...
+
+    getResults(new EnumSearch("Pacific"));
+    getResults(new EnumSearch("Mountain"));
+    getResults(new EnumSearch("Central"));
+    getResults(new EnumSearch("Eastern"));
+    // intentionally make sure a bogus term just returns no results
+    getResults(new EnumSearch("Atlantic"));
+  }
+
+  private void getResults(final EnumSearch search) {
```
jdgarrett, line 201: Wouldn't both of these queries use the same index because of the custom constraints?  Seems like if for some reason some entries were being left out of the result it wouldn't be caught.  Maybe there should be an assertion for the number of results in each set as well.

jdgarrett (commented):
This looks good to me, I just have one concern/question about the integration test.

test/src/test/java/org/locationtech/geowave/test/basic/GeoWaveEnumIndexIT.java
rfecher: good catch!  I intended to leave the first query collecting expected results as unconstrained and just apply a manual filter on the full dataset, but I guess I goofed.

PR #1805, opened 2021-04-26
GEOWAVE-1804: spatial binning strategies

rfecher:
Signed-off-by: Rich Fecher <richard.fecher@maxar.com>


examples/java-api/src/main/java/org/locationtech/geowave/examples/stats/SpatialBinningStatisticExample.java
```diff
@@ -0,0 +1,309 @@
+package org.locationtech.geowave.examples.stats;
```
jdgarrett, line 1: Missing header

jdgarrett (changes requested):
Looks good to me, there were just a couple of things that I thought should be addressed.

PR #1813, opened 2021-05-28
GEOWAVE-1812: updated major dependencies to current versions

rfecher:


extensions/datastores/hbase/coprocessors/src/main/java/org/locationtech/geowave/datastore/hbase/coprocessors/HBaseBulkDeleteEndpoint.java
```diff
@@ -326,4 +335,36 @@ public void start(final CoprocessorEnvironment env) throws IOException {
   public void stop(final CoprocessorEnvironment env) throws IOException {
     // nothing to do
   }
+
+  public static void main(final String[] args) throws Exception {
```
jdgarrett, line 339: This looks like left over test code that should probably be removed.

python/src/main/python/pygw/store/cassandra/options.py
```diff
@@ -86,3 +103,54 @@ def get_replication_factor(self):
             The number of replicas to use when creating a new keyspace.
         """
         return self._base_options.getReplicationFactor()
+
+    def set_gc_grace_seconds(self, gc_grace_seconds):
```
jdgarrett, line 107: These gc_grace_seconds methods aren't being covered in the test.

test/src/test/java/org/locationtech/geowave/test/landsat/CustomCRSLandsatIT.java
```diff
@@ -117,6 +118,10 @@ protected void processParameters(final OperationParams params) throws Exception
 
   @BeforeClass
   public static void startTimer() {
+
+    // GDAL version >= 2.3.0
+    System.loadLibrary("gdalalljni");
```
jdgarrett, line 123: I am curious as to why gdal needs to be explicitly loaded here.  Does that mean that third party developers need to explicitly load gdal when using those capabilities within GeoWave?

jdgarrett (changes requested):

extensions/adapters/raster/src/main/java/org/locationtech/geowave/adapter/raster/plugin/gdal/InstallGdal.java
```diff
@@ -73,13 +73,13 @@ private static void install(final File gdalDir) throws IOException {
       gdalEnv = DEFAULT_BASE;
     }
     if (isWindows()) {
-      file = "gdal-1.9.2-MSVC2010-x64.zip";
-      url = new URL(gdalEnv + "/windows/MSVC2010/" + file);
+      file = "win-x64-gdal204.zip";
+      url = new URL(gdalEnv + "/windows/MSVC2017/" + file);
     } else if (isMac()) {
       file = "gdal-1.9.2_macOSX.zip";
```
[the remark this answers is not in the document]
rfecher: it _may_ be best to stick with homebrew on the mac, although I spent awhile making sure this worked at least under windows (for ease with our standalone installer) and ubuntu 20.04 (for CI tests), although I don't think it works across the board on linux even.  I don't have any options for testing on a mac so I'd not be the best person to try to get it working for mac osx.  The older version seemed very specific with an older GDAL install which really pushed us toward integrating the install with our CLI for ease of use, but this should be good for GDAL 2.2+ (any recent version) so at some point we could just decide to ditch the install gdal CLI.  Also, technically, GeoTools actually picks up on the older GDAL version, but our tests don't pass with some method not found exceptions (so its probably not very functional, might work for some things though for GeoTools to even attempt to use it?).

extensions/datastores/accumulo/src/main/java/org/locationtech/geowave/datastore/accumulo/operations/AccumuloOperations.java
```diff
@@ -1710,4 +1746,18 @@ public void deleteRowsFromDataIndex(final byte[][] rows, final short adapterId)
       }
     }
   }
+
+  @Override
+  public void close() {
```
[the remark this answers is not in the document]
rfecher: its similar to how RocksDB works, the `DataStore` is closeable as well and additionally we do call it from GeoTools' DataStore's dispose() method.  In both cases we use a static cache for connections so there is no sense in opening and closing anything on a per instance basis, just at the end of the JVMs usage (but when the JVM exits I'd bet resources are released as well)

extensions/datastores/hbase/coprocessors/src/main/java/org/locationtech/geowave/datastore/hbase/coprocessors/HBaseBulkDeleteEndpoint.java
rfecher: good catch

python/src/main/python/pygw/store/cassandra/options.py
rfecher: good catch

test/src/test/java/org/locationtech/geowave/test/landsat/CustomCRSLandsatIT.java
rfecher: nope, another left-over from fighting the tests, I'll delete

PR #1825, opened 2021-06-14
GEOWAVE-1824: fixed rocksdb lockfile contention

rfecher:


extensions/datastores/rocksdb/src/test/java/org/locationtech/geowave/datastore/rocksdb/RocksDBTestPersistableRegistry.java
```diff
@@ -0,0 +1,13 @@
+package org.locationtech.geowave.datastore.rocksdb;
```
jdgarrett, line 1: Missing header

jdgarrett (approved):
Looks good to me, just missing one header comment

PR #1873, opened 2022-03-04
Suppress multiple binding warnings for installer

mzagorskirs:
Testing the log4j updates in the installer revealed the need to exclude the new log4j version to suppress the multiple binding warnings when using the installed version of GeoWave.


mzagorskirs:
OBE - closing

PR #1932, opened 2026-09-20
Take json-lib out of the core artifacts

rfecher:
> Stacked on #1931.

json-lib 2.4 is from 2010 and drags commons-beanutils, commons-collections **3.2.1**, **commons-lang 2.6** and ezmorph behind it. Removing it from `core/index` and `core/cli` drops ezmorph and json-lib outright and moves commons-collections to 3.2.2, the version without the `InvokerTransformer` deserialization gadget.

**`IndexMetaData.toJSONObject` has no callers anywhere in the tree** and could not be kept without keeping json-lib, since the return type is json-lib's own. It goes, along with its three implementations. This is an API break for anyone implementing `IndexMetaData` outside GeoWave — I checked the downstream consumer we care about and it does not.

`ValueConverter`'s json-lib check becomes a `JsonNode` check, preserving the intent: already-parsed JSON passes through instead of being stringified. `core/cli` now declares commons-beanutils explicitly — it was compiling against a transitive of json-lib — and picks up 1.11.0. `services/client` declared json-lib without using it.

**Seventeen files were compiling against commons-lang 2.6 with nothing declaring it either**, across `core/store`, `core/geotime`, `core/mapreduce`, `accumulo`, `adapters/vector`, two format modules, `services/rest` and `analytics/spark`. They move to lang3, which the rest of the tree already uses and which has the same `ArrayUtils` and `StringUtils` members. Without this the tree does not compile at this commit — that is why it is here rather than later in the series.

`extensions/cli/geoserver` is the last consumer and gets an explicit declaration so it keeps building; it moves to Jackson in #1937, after which json-lib leaves the build entirely.


rfecher:
Superseded by the PR above: this one was auto-closed when its base branch `dependency-cleanup` was deleted on merge of #1931, and GitHub will not reopen a PR whose base is gone. Same branch, same commit, retargeted at master.

PR #1953, opened 2026-09-21
Java 21: maven.compiler.release, Spark 4 / Scala 2.13, URLStreamHandlerProvider

rfecher:
These three cannot be separated. enforceBytecodeVersion bans Spark 4 outright
while the target is 1.8 -- it is class-file major 61 -- and the acceptance
criterion for the flip is the IT matrix green on 21 with no --add-opens, which
nothing older than Spark 4 can meet: 3.1.1 dies on scala.Product.$init$, 3.5.7
on IllegalAccessError for sun.nio.ch.DirectBuffer. And the java.net.URL
reflection cannot be replaced before the flip either, because
java.net.spi.URLStreamHandlerProvider is Java 9+.

  maven.compiler.target 1.8 -> 21, and the compiler plugin now takes <release>
  rather than <source>/<target>, so the Java 21 API surface is checked and not
  just the class-file version. enforceBytecodeVersion reads the same property,
  so its ceiling moves with it.

  Spark 3.1.1 -> 4.0.1 and Scala 2.12.13 -> 2.13.16, across six poms plus
  kafka, scala-collection-compat and s3mock. Two source changes: SparkSession
  in Spark 4 no longer overrides close(), so it inherits Closeable's throws
  IOException -- six call sites now use stop(), which both versions define as
  the same thing and which throws nothing. The Ordering.Double change landed
  separately in #1951.

  slf4j 1.7.30 -> 2.0.16 and log4j 2.17.1 -> 2.24.3, log4j-slf4j-impl ->
  log4j-slf4j2-impl. This is not housekeeping: Spark 4 brings log4j-slf4j2-impl
  and GeoWave brought log4j-slf4j-impl, both of which ship
  org.apache.logging.slf4j.Log4jLoggerFactory with different constructors. The
  result was NoSuchMethodError: Log4jLoggerFactory.<init>() the first time
  anything initialised logging.

  URLStreamHandlerProvider replaces three copies of a routine that called
  URL.setURLStreamHandlerFactory and then, because the JDK permits that once per
  JVM, reflected into URL.factory and URL.streamHandlerLock to install a second
  handler behind the first. streamHandlerLock was deleted in JDK 18, so no flag
  brings that back. Providers need no chaining -- the JDK asks each in turn --
  so the two delegating factories go with the reflection, as does
  URLClassloaderUtils.verifyProtocol, which existed to install a handler before
  new URL() could be trusted.

Three things found on the way that are worth knowing.

A provider must not hold a logger. The JDK builds providers from inside
URL.<init>, so the first version of this class failed its static init on the
slf4j clash above and turned every new URL() in the JVM, of any protocol, into
a ServiceConfigurationError. The class now touches nothing that can fail to
initialise, and says so.

The provider answers s3 and hdfs and nothing else. The old
HdfsUrlStreamHandlerFactory delegated every unmatched protocol to Hadoop's
FsUrlStreamHandlerFactory, which has entries for schemes including http. As the
JVM-wide factory, that was live exposure.

Provider lookup is system-classloader only:
ServiceLoader.load(URLStreamHandlerProvider.class,
ClassLoader.getSystemClassLoader()). Code under a child-first loader cannot see
it, which is why QueryFilterIterator still installs a factory directly -- it
runs inside an Accumulo tablet server. A passing unit test does not speak for a
lane, and the s3 and hdfs ingest ITs are what will.

Two smaller items. javac 21's bytecode let SpotBugs see that two null checks in
MemoryStoreUtils' visibility parser are dead: processTerm has three exits, one
that throws, one that returns a new ChildExpression and one that returns a
value it has already thrown on if null, so it never returns null, and neither
does parse_. The "()" case still reports "empty term" from processTerm, which
is the message it already produced. And CI moves to 21 for unit tests,
integration tests and publishing; the JDK 21 canary goes, having become the
build. No canary replaces it: the next LTS is blocked on HADOOP-19486, so a
JDK 25 lane would be permanently red, which is the kind of gate people learn to
click past.

Built on JDK 21 with zero --add-opens: 56 modules, SpotBugs clean. The only
local test failures are rocksdbjni and embedded redis, whose native binaries
are x86_64-only and cannot load on Apple Silicon at any JDK version.

Tests and IP Check also run on push to the 3.0 branch now; Publish stays on
master, so a 3.0.0-SNAPSHOT has no path to the Portal.


rfecher:
Follow-up detail on `Never report a zero-length split that still has ranges to scan` (7ef481b2a0d), verified after the commit message was written:

`spark.hadoopRDD.ignoreEmptySplits` **defaults `false` in Spark 3.1.1 and `true` in Spark 4.0.1**. Read out of the two jars rather than the docs — at the `createWithDefault` in `org.apache.spark.internal.config.package$`, Spark 3.1.1 pushes `iconst_0` and Spark 4.0.1 pushes `iconst_1`.

So `GeoWaveInputSplit.getLength()` returning zero for a split that still has ranges to scan is an old bug that was never reachable: Spark 3 created a partition for every split regardless of its length. This branch is the first thing to run GeoWave against a Spark that acts on that value.

PR #1954, opened 2026-09-22
Name filesystem store files with base32hex, which is single-case

rfecher:
## ⚠️ Breaking change: every existing filesystem store has to be re-ingested

This changes the on-disk format of the `filesystem` datastore. There is no in-place migration — see
the new "Changes That Migration Cannot Handle" section in
`docs/content/userguide/101-migrating.adoc`.

## The bug

Every metadata entry and every data row in a filesystem store is one file whose name encodes
that entry's key, and each index partition is a directory named the same way. Both used URL-safe
base64 (`ByteArrayUtils.byteArrayToString`), whose alphabet uses upper and lower case to mean
different things: `AA.bin` and `aA.bin` are the keys `0x00` and `0x68`.

On a case-insensitive filesystem — the default for macOS APFS, the norm for Windows NTFS — those
are one file. `Files.write(..., CREATE, TRUNCATE_EXISTING)` then overwrote the existing entry
**while keeping the existing name**, so the entry just written was invisible under its own key and
the entry it landed on was destroyed.

**104 of the 256 one-byte keys collide this way**, so a store on such a filesystem was losing
metadata and rows routinely, not in a corner case.

## How it surfaced

`CustomCRSKDERasterResizeIT` fails on macOS with `NullPointerException: ... because "crs" is null`
at `GeoWaveRasterReader.getOriginalEnvelope:332`. The resize output type had no adapter-to-index
mapping: `indexMappingStore.addAdapterIndexMapping(...)` was followed by an immediate
`getIndicesForAdapter(2)` returning zero rows, while the same sequence for adapter 125 moments
earlier returned one. The second write had landed on the first one's file.

## The fix

Names are now **unpadded base32hex** (RFC 4648 extended hex alphabet, `0-9A-V`). Two properties
matter:

- **Single-case**, so distinct keys cannot collide however the filesystem folds case.
- **Order-preserving** — which base64 was *not*. The read path treats a directory listing as a
  sorted key space (`getSortedSet` takes a `tailSet`/`headSet` over byte-range bounds, metadata
  prefix scans run `[prefix, nextPrefix)`), so encoded names now sort the way the keys they encode
  do.

base32hex over lowercase hex because it is 1.6× the key size against hex's 2×, which matters under
a 255-character name limit: the longest usable key is ~156 bytes rather than ~125. It is still a
reduction from base64's ~188, which the migration note calls out.

`decodeKey` rejects a name it cannot read rather than guessing at it, so pointing a new build at an
old store fails with an error naming the file instead of silently misreading it.

The partition **directory** name (`FileSystemDataFormatter.getPartitionDirectoryName`) had the
identical defect and is fixed the same way — leaving it would have left the store corrupting.

## Two nearby defects

Not the cause; they are what made this hard to read.

- `GeoWaveOutputFormat.getIndexWriter` cached a null writer for a type with no indices, and
  `close()` iterates that cache — so `"Cannot write to index ..."`, the message `write()` had
  already built, was buried under an NPE out of `close()`.
- `AdapterStoreWrapper.getAdapter(Short)` called `asInternalAdapter` on a missing adapter, turning
  the interface's documented *"null if it doesn't exist"* into an NPE deep inside a Spark task.

## Tests

New `FileSystemUtilsKeyEncodingTest` round-trips keys, asserts distinct keys never share a name
when case is ignored, and asserts encoded names sort like their raw keys across differing lengths
(random keys, every prefix of a key, and `0x00`/`0xff` runs). Against the old base64 encoding 3 of
its 5 cases fail, the first with
`'AA.bin' and 'aA.bin' are the same file on a case-insensitive filesystem`.

## Verification

The code here is byte-identical to the `master` companion branch (the branches differ only by the
`2.0.2-SNAPSHOT` → `3.0.0-SNAPSHOT` pom bump and the version named in the migration note), and
identical to what was run on top of #1953, so the three verification passes below cover it between
them:

- On the #1953 stack (`3.0` + Java 21), full `-P filesystem-it` lane: **158 tests, 0 failures, 0
  errors**, 23 skipped (all `GeoServerIT` plus `IngestIT.sparkToGW` — environment-dependent,
  unrelated). `CustomCRSKDERasterResizeIT` passes there, having previously NPE'd.
- On this branch's code at `maven.compiler.target 1.8`: filesystem module compiles, its 5 unit
  tests pass.
- `formatter:validate` clean.

## Merge-order note

#1953 also adds the `==== Changes That Migration Cannot Handle` heading to
`101-migrating.adoc`, for the redis format break. Whichever of the two lands second will need a
one-line conflict resolution on that heading; the two notes sit side by side underneath it.


rfecher:
Companion PR for the JDK 8 line: #1955 (same fix, based on `master`). The two carry byte-identical Java; they differ only in the pom version and the version named in the migration note.

